PR #1, opened 2020-02-05
Unreal mode

Maxez:
plo starts in 'big unreal mode'. Kernel gets loaded to high memory (default kernel base address changed to c0110000).


ia32/a20.c
```diff
@@ -0,0 +1,123 @@
...
+
+					if (!wait_kbd_status(1)) {
+						/* Now test the A20 line */
+						if (a20_test())
+							return 1;
+					}
+				}
+			}
+		}
+	}
```
agkaminski: Too much _if_ nesting

pawelpisarczyk: I think that these if's could be removed. Wait function is blocking.

PR #3, opened 2020-02-24
Fixed IA32 Unreal Mode

Maxez:


ia32/plo.c
```diff
@@ -213,34 +213,52 @@ void plo_unlimitfs(void)
...
+	lgdt 0
 
-	; Load gdt register
-	mov eax, #SYSPAGE_SEG
-	lgdt [eax]
-	
 	; Switch to pmode by setting pmode bit
 	mov eax, cr0
-	or al, 1
+	inc eax
```
agkaminski: Are we 100% sure LSB is not set already in this case?

Maxez: Yes, plo starts in the real mode with cr0 LSB zeroed out. Confirmed by Paweł.

PR #6, opened 2020-06-22
ia32: Added 'copy' command for copying data between devices

Maxez:


ia32/cmd.c
```diff
@@ -155,6 +157,39 @@ char *cmd_getnext(char *line, unsigned int *pos, char *blanks, char *cites, char
...
+/* Function parses file name / LBA and returns its handle */
+static s32 cmd_gethandle(unsigned int dn, char *line, unsigned int *pos, char *blanks, char *cites, char *word, unsigned int len)
+{
+	if (cmd_getnext(line, pos, blanks, cites, word, len) == NULL)
+		return ERR_ARG;
+
+	if (*word == '\0')
+		return ERR_ARG;
+
+	return (dn >= PDN_COM1 && dn <= PDN_COM2) ? phfs_open(devices[dn].pdn, word, 0) : (s32)plostd_atol(word);
```
Maxez, line 189: This line could be replaced by just `return phfs_open(devices[dn].pdn, word, 0);` if disk_open() was implemented like so:
```
s32 disk_open(u16 dn, char *name, u32 flags)
{
    return plostd_atol(word);
}
```

PR #12, opened 2020-08-29
armv7m7: add kernel/file load and dump/write cmds

Cinek28:


pawelpisarczyk:
Marcin, pleas correct indentation and be consequent is using tabs and space. Just in case - we agreed to use tabs to indent the code.

Cinek28:
Fixed indentation in commited files

PR #15, opened 2020-09-25
cmd.c: code optimization

kemonats:


cmd.c
```diff
@@ -151,29 +151,24 @@ void cmd_cmd(char *s)
...
-		}
-		if (!blfl)
+	char *p;
+
+	line += *pos;
+	while (*line) {
+		for (p = blanks; *p && *line != *p; p++);
+		if (!*p)
 			break;
+		line++;
```
agkaminski, line 171: I think that use of line[*pos] would be simpler overall

agkaminski (commented):
Only one small remark. Rest of change looks good to me

kemonats:
For me, the pointer version is simpler and a few bytes shorter, but this can be changed. Should I change it?

agkaminski:
I've meant someting along these lines:

`void cmd_skipblanks(char *line, unsigned int *pos, char *blanks)
{
	char *p;

	while (*line) {
		for (p = blanks; *p && *(line + *pos) != *p; p++);
		if (!*p)
			break;
		(*pos)++;
	}

	return;
}`

Two lines less ;)

kemonats:
I changed to your version.

HBuczynski:
Your suggestions have been applied to a new version of plo.
PR is closed.

PR #26, opened 2021-02-05
Enable jtag/debugger and some fixes

gerard5:
This commit is analogous to https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/178 because of the same procedure.

* The debugger functionality has been checked against to use J-Link (jtag / swd) and [pyOCD](https://github.com/pyocd/pyOCD) (using CMSIS-DAP available on the mimixrt1064-evk board).
* Also a minor update in the comment about clock freq.

### DIFF
--- a/armv7m7-imxrt106x/imxrt.c
+++ b/armv7m7-imxrt106x/imxrt.c
@@ -1898,6 +1898,9 @@ void _imxrt_init(void)
 	*(imxrt_common.ccm + ccm_ccgr5) = 0xf00f330f;
 	*(imxrt_common.ccm + ccm_ccgr6) = 0x00fc0f0f;
 
+	imxrt_dataSyncBarrier();
+	imxrt_dataInstrBarrier();
+
 	/* Remain in run mode on wfi */
 	_imxrt_ccmSetMode(0);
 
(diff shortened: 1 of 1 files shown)


armv7m7-imxrt106x/low.c
```diff
@@ -317,15 +317,6 @@ int low_launch(void)
...
 
-	/* Give the LPUART transmitters some time */
-	timer_wait(100, TIMER_EXPIRE, NULL, 0);
-
-	/* Tidy up */
-	serial_done();
-	timer_done();
-
-	low_done();
-
```
agkaminski, line 328: My mistake, I think it would be better to leave this as is, remove changes from cmd_go and make low_launch in imxrt117x analogous to this one.

agkaminski (commented):
I've missed changes in low_launch in imxrt106x when working on imxrt117x. I think we should revert changes in cmd_go. 

Waiting for confirmation from @Cinek28 that this PR solves the problem.

armv7m7-imxrt106x/low.c
gerard5: Sounds good.

Cinek28:
I checked this PR on imxrt106x board and everything works good now. I can use jtag in plo and after kernel launch.
Every peripheral used in project is working as it should. Approved.

PR #33, opened 2021-03-04
imxrt: add separate uart console flag for plo

Cinek28:
The flag UART_CONSOLE_PLO is optional. Should be
used when it is necessary to split up system console
with one for plo. If flag is not defined UART_CONSOLE
takes effect.


agkaminski:
Can You connect it to some JIRA task?

Cinek28:
Created and added jira task label

PR #39, opened 2021-03-23
usb driver, phfs interafce and modification of phoenixd & msg

HBuczynski:
The following PR contains:
- USB client driver,
- CDC driver,
- changes in phoenixd & msg allow to use different phfs,
- phfs-serial - interface to exchange data via serial with host app **phoenixd**,
- phfs-usb - interface to exchange data via usb with host app **phoenixd**,
- clean up device initialization,
- add lib module with doubly-linked list,

The code works as expected on **zynq7000** and **imxrt106x** ( https://github.com/phoenix-rtos/plo/pull/38 has to be added).
To merge this PR, the code should be tested on **imxrt117x**. @MaciejPurski, can you verify it on hardware?

### DIFF
--- a/armv7m7-imxrt106x/Makefile
+++ b/armv7m7-imxrt106x/Makefile
@@ -9,8 +9,8 @@
 INIT_RAM = 0x0
 INIT_FLASH = 0x70000000
 
-BSS=-Tbss=20011000
-LDFLAGS+=--defsym=plo_bss=0x20011000
+BSS=-Tbss=20010000
+LDFLAGS+=--defsym=plo_bss=0x20010000
 
 LDFLAGS:=$(filter-out -Tbss% , $(LDFLAGS))
 LDFLAGS:=$(filter-out -Tdata% , $(LDFLAGS))
@@ -20,4 +20,4 @@ CFLAGS+= -mfloat-abi=soft
 
 GCCLIB := $(shell $(CC) $(CFLAGS) -print-libgcc-file-name)
 
-OBJS += $(addprefix $(PREFIX_O)$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)/, _startup.o low.o imxrt.o serial.o cmd.o timer.o flashdrv.o rom_api.o flashcfg.o phfs-flash.o syspage.o)
+OBJS += $(addprefix $(PREFIX_O)$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)/, _startup.o low.o imxrt.o serial.o cmd.o timer.o flashdrv.o rom_api.o flashcfg.o phfs-flash.o syspage.o usbphy.o)
--- a/devices/usbc/cdc-client.c
+++ b/devices/usbc/cdc-client.c
@@ -1,0 +1,421 @@
+/*
+ * Phoenix-RTOS
+ *
+ * phoenix-rtos-loader
+ *
+ * cdc - USB Communication Device Class
+ *
+ * Copyright 2019, 2021 Phoenix Systems
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+
+#include "usbclient.h"
+#include "cdc-client.h"
+
+#include "list.h"
+#include "config.h"
+#include "../errors.h"
+
+
+struct {
+	usb_desc_list_t *descList;
+
+	usb_desc_list_t dev;
+	usb_desc_list_t conf;
+
+	struct {
+		usb_desc_list_t iad;
+
+		usb_desc_list_t comIface;
+		usb_desc_list_t header;
+		usb_desc_list_t call;
+		usb_desc_list_t acm;
+		usb_desc_list_t unio;
+		usb_desc_list_t comEp;
+
+		usb_desc_list_t dataIface;
+		usb_desc_list_t dataEpOUT;
+		usb_desc_list_t dataEpIN;
+	} ports[PHFS_ACM_PORTS_NB];
+
+	usb_desc_list_t str0;
+	usb_desc_list_t strman;
+	usb_desc_list_t strprod;
+} cdc_common;
+
+
+/* Device descriptor */
+const usb_device_desc_t dDev = {
+	.bLength = sizeof(usb_device_desc_t),
+	.bDescriptorType = USB_DESC_DEVICE,
+	.bcdUSB = 0x0002,
+	.bDeviceClass = 0x0,
+	.bDeviceSubClass = 0,
+	.bDeviceProtocol = 0,
+	.bMaxPacketSize0 = 64,
+	.idVendor = 0x16f9,
+	.idProduct = 0x0003,
+	.bcdDevice = 0x0200,
+	.iManufacturer = 1,
+	.iProduct = 2,
+	.iSerialNumber = 0,
+	.bNumConfigurations = 1
+};
+
+
+/* Configuration descriptor */
+const usb_configuration_desc_t dConfig = {
+	.bLength = 9, .bDescriptorType = USB_DESC_CONFIG,
+	.wTotalLength = sizeof(usb_configuration_desc_t) + (sizeof(usb_interface_desc_t) + sizeof(usb_desc_cdc_header_t) + sizeof(usb_desc_cdc_call_t)
+				+ sizeof(usb_desc_cdc_acm_t) + sizeof(usb_desc_cdc_union_t) + sizeof(usb_endpoint_desc_t) + sizeof(usb_interface_desc_t)
+				+ sizeof(usb_endpoint_desc_t) + sizeof(usb_endpoint_desc_t)) * PHFS_ACM_PORTS_NB,
+	.bNumInterfaces = 2 * PHFS_ACM_PORTS_NB,
+	.bConfigurationValue = 1,
+	.iConfiguration = 0,
+	.bmAttributes = 0xc0,
+	.bMaxPower = 5
+};
+
+
+const usb_interface_association_desc_t dIad[]= {
+	{
+		.bLength = 0x08,
+		.bDescriptorType = 0x0b,
+		.bFirstInterface = 0,
+		.bInterfaceCount = 2,
+		.bFunctionClass = 2,
+		.bFunctionSubClass = 2,
+		.bFunctionProtocol = 1,
+		.iFunction = 0,
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 0x08,
+		.bDescriptorType = 0x0b,
+		.bFirstInterface = 0x2,
+		.bInterfaceCount = 2,
+		.bFunctionClass = 2,
+		.bFunctionSubClass = 2,
+		.bFunctionProtocol = 1,
+		.iFunction = 0,
+	}
+#endif
+};
+
+/* Communication Interface Descriptor */
+const usb_interface_desc_t dComIntf[] = {
+	{
+		.bLength = 9,
+		.bDescriptorType = USB_DESC_INTERFACE,
+		.bInterfaceNumber = 0,
+		.bAlternateSetting = 0,
+		.bNumEndpoints = 1,
+		.bInterfaceClass = 0x02,
+		.bInterfaceSubClass = 0x02,
+		.bInterfaceProtocol = 0x00,
+		.iInterface = 4
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 9,
+		.bDescriptorType = USB_DESC_INTERFACE,
+		.bInterfaceNumber = 2,
+		.bAlternateSetting = 0,
+		.bNumEndpoints = 1,
+		.bInterfaceClass = 0x02,
+		.bInterfaceSubClass = 0x02,
+		.bInterfaceProtocol = 0x00,
+		.iInterface = 4
+	}
+#endif
+};
+
+
+const usb_desc_cdc_header_t dHeader[] = {
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0,
+		.bcdCDC = 0x0110
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0,
+		.bcdCDC = 0x0110
+	}
+#endif
+};
+
+
+const usb_desc_cdc_call_t dCall[] = {
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x01,
+		.bmCapabilities = 0x01,
+		.bDataInterface = 0x1
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x01,
+		.bmCapabilities = 0x01,
+		.bDataInterface = 0x1
+	}
+#endif
+};
+
+
+const usb_desc_cdc_acm_t dAcm[] = {
+	{
+		.bLength = 4,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x02,
+		.bmCapabilities = 0x03
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 4,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x02,
+		.bmCapabilities = 0x03
+	}
+#endif
+};
+
+
+const usb_desc_cdc_union_t dUnion[] = {
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x06,
+		.bControlInterface = 0x0,
+		.bSubordinateInterface = 0x1
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 5,
+		.bType = USB_DESC_TYPE_CDC_CS_INTERFACE,
+		.bSubType = 0x06,
+		.bControlInterface = 0x2,
+		.bSubordinateInterface = 0x3
+	}
+#endif
+};
+
+
+/* Communication Interrupt Endpoint IN */
+const usb_endpoint_desc_t dComEp[] = {
+	{
+		.bLength = 7,
+		.bDescriptorType = USB_DESC_ENDPOINT,
+		.bEndpointAddress = 0x80 | endpt_irq_acm0, /* direction IN */
+		.bmAttributes = 0x03,
+		.wMaxPacketSize = 0x20,
+		.bInterval = 0x08
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 7,
+		.bDescriptorType = USB_DESC_ENDPOINT,
+		.bEndpointAddress = 0x80 | endpt_irq_acm1, /* direction IN */
+		.bmAttributes = 0x03,
+		.wMaxPacketSize = 0x20,
+		.bInterval = 0x08
+	}
+#endif
+};
+
+
+/* CDC Data Interface Descriptor */
+const usb_interface_desc_t dDataIntf[] = {
+	{
+		.bLength = 9,
+		.bDescriptorType = USB_DESC_INTERFACE,
+		.bInterfaceNumber = 1,
+		.bAlternateSetting = 0,
+		.bNumEndpoints = 2,
+		.bInterfaceClass = 0x0a,
+		.bInterfaceSubClass = 0x00,
+		.bInterfaceProtocol = 0x00,
+		.iInterface = 0
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 9,
+		.bDescriptorType = USB_DESC_INTERFACE,
+		.bInterfaceNumber = 3,
+		.bAlternateSetting = 0,
+		.bNumEndpoints = 2,
+		.bInterfaceClass = 0x0a,
+		.bInterfaceSubClass = 0x00,
+		.bInterfaceProtocol = 0x00,
+		.iInterface = 0
+	}
+#endif
+};
+
+
+/* Data Bulk Endpoint OUT */
+const usb_endpoint_desc_t dEpOUT[] = {
+	{
+		.bLength = 7,
+		.bDescriptorType = USB_DESC_ENDPOINT,
+		.bEndpointAddress = endpt_bulk_acm0, /* direction OUT */
+		.bmAttributes = 0x02,
+		.wMaxPacketSize = 0x0200,
+		.bInterval = 0
+	},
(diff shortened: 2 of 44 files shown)


devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+	return 1;
+}
+
+
+int ctrl_lfIrq(void)
+{
+	if ((*(ctrl_common.dc->base + usbsts) & 1 << 6)) {
+
+		*(ctrl_common.dc->base + endptsetupstat) = *(ctrl_common.dc->base + endptsetupstat);
+		*(ctrl_common.dc->base + endptcomplete) = *(ctrl_common.dc->base + endptcomplete);
```
kemonats, line 329: It is correct?

devices/usbc/desc.c
```diff
@@ -0,0 +1,353 @@
...
+static void desc_defaultSetup(const usb_setup_packet_t *setup)
+{
+	u32 fsz;
+
+	if (*(u32 *)setup == 0xdeadc0de) {
+		desc_common.dc->op = DC_OP_EXIT;
+	}
+	else {
+		fsz = setup->wValue << 16;
+		fsz |= setup->wIndex;
```
kemonats, line 248: `fsz` is assigned  a value that is never used.

devices/usbc/client.c
```diff
@@ -0,0 +1,167 @@
...
+
+	while ((imx_common.dc.op != DC_OP_RCV_ENDP0) && (imx_common.dc.op != DC_OP_RCV_ERR))
+
+	if (imx_common.dc.op != DC_OP_RCV_ERR) {
+		res = imx_common.data.endpts[0].buf[USB_ENDPT_DIR_OUT].len;
+		low_memcpy(data, (const char *)imx_common.data.endpts[0].buf[USB_ENDPT_DIR_OUT].buffer, res); /* copy data to buffer */
+		imx_common.dc.op = DC_OP_NONE;
+
+		ctrl_execTransfer(0, (u32)imx_common.data.endpts[0].buf[USB_ENDPT_DIR_IN].buffer, 0, USB_ENDPT_DIR_IN); /* ACK */
+	}
```
kemonats, line 61: Inner `if` condition is always `true`.

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+	return ERR_NONE;
+}
+
+
+dtd_t *ctrl_execTransfer(int endpt, u32 paddr, u32 sz, int dir)
+{
+	int shift;
+	u32 offs;
+	volatile dtd_t *dtd;
+
```
kemonats, line 262:
```suggestion

```
unnecessary tabulation

devices/usbc/cdc-client.c
```diff
@@ -0,0 +1,421 @@
...
+	} ports[PHFS_ACM_PORTS_NB];
+
+	usb_desc_list_t str0;
+	usb_desc_list_t strman;
+	usb_desc_list_t strprod;
+} cdc_common;
+
+
+/* Device descriptor */
+const usb_device_desc_t dDev = {
```
kemonats, line 53: trailing whitespace in the lines 53-331

devices/usbc/cdc-client.c
```diff
@@ -0,0 +1,421 @@
...
+const usb_interface_association_desc_t dIad[]= {
+	{
+		.bLength = 0x08,
+		.bDescriptorType = 0x0b,
+		.bFirstInterface = 0,
+		.bInterfaceCount = 2,
+		.bFunctionClass = 2,
+		.bFunctionSubClass = 2,
+		.bFunctionProtocol = 1,
+		.iFunction =0,
```
kemonats, line 94:
```suggestion
		.iFunction = 0,
```

devices/usbc/cdc-client.c
```diff
@@ -0,0 +1,421 @@
...
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 0x08,
+		.bDescriptorType = 0x0b,
+		.bFirstInterface = 0x2,
+		.bInterfaceCount = 2,
+		.bFunctionClass = 2,
+		.bFunctionSubClass = 2,
+		.bFunctionProtocol = 1,
+		.iFunction =0,
```
kemonats, line 105:
```suggestion
		.iFunction = 0,
```

devices/usbc/ctrl.c
HBuczynski: done

devices/usbc/cdc-client.c
HBuczynski: Thanks

devices/usbc/cdc-client.c
HBuczynski: Thanks

devices/usbc/cdc-client.c
HBuczynski: Thanks, whitespaces were removed.

msg.c
```diff
@@ -26,136 +26,130 @@
...
-    csum += msg_getseq(msg);
-    return csum;
+	u32 k;
+	u16 csum;
+
+	csum = 0;
+	for (k = 0; k < MSG_HDRSZ + msg_getlen(msg); k++) {
+		if (k >= sizeof(msg->csum))
+			csum += *((u8 *)msg + k);
+	}
```
kemonats, line 36:
```suggestion
	for (k = sizeof(msg->csum); k < MSG_HDRSZ + msg_getlen(msg); k++) {
		csum += *((u8 *)msg + k);
	}
```
I was going to create PR, but if you are changing this file then by the way this change. 
It will be faster.

msg.c
```diff
@@ -26,136 +26,130 @@
...
+			else
+				cs[1] = MSG_ESCESC;
+			if ((res = msg->write(pn, (u8 *)cs, 2)) < 0)
+				return res;
+		}
+		else {
+			if ((res = msg->write(pn, (u8 *)&p[k], 1)) < 0)
+				return res;
+		}
+	}
```
kemonats, line 68:
```suggestion
	u8 cs[2];
	u16 k;
	int res;

	/* Frame start */
	cs[0] = MSG_MARK;
	if ((res = msg->write(pn, cs, 1)) < 0)
		return res;

	for (k = 0; k < MSG_HDRSZ + msg_getlen(msg); k++) {
		if ((p[k] == MSG_MARK) || (p[k] == MSG_ESC)) {
			cs[0] = MSG_ESC;
			cs[1] = p[k] == MSG_MARK ? MSG_ESCMARK : MSG_ESCESC;
			if ((res = msg->write(pn, cs, 2)) < 0)
				return res;
		}
		else {
			if ((res = msg->write(pn, &p[k], 1)) < 0)
				return res;
		}
	}
```
And also by the way:
- `cs`  as `u8` type instead of `char`
- `pk` is of type `u8 *`

devices/usbc/ctrl.c
HBuczynski: This action is performed when USB reset state is received. 
Registers `endptsetupstat` and `endptcomplete` contain information about active transfers on specific endpoints. After reset, all active transfers should be cleared. 
According to documentation: _Writing one clears the corresponding bit in this register_. For this reason the same value is written to register to clear its state.

devices/usbc/desc.c
HBuczynski: Thanks, it was removed.

devices/usbc/client.c
HBuczynski: I think that removing `(imx_common.dc.op != DC_OP_RCV_ERR)` from while loop is a better way. Then the 'if' condition makes sense.

devices/usbc/client.h
```diff
@@ -0,0 +1,179 @@
...
+#define DTD_ERROR(dtd)      (dtd->dtd_token & (0xd << 3))
+#define DTD_ACTIVE(dtd)     (dtd->dtd_token & (1 << 7))
+
+
+/* device controller structures */
+
+/* data transfer descriptor */
+typedef struct _dtd_t {
+    u32 dtd_next;
+    u32 dtd_token;
```
kemonats, line 41: In the files in directory `devices/usbc:
```
client.h
usbclient.h
usb.h
```
correct indentation.

msg.c
HBuczynski: Thanks for your note, it is obviously correct. 
However, this PR is related to add USB driver and modules associated with it and its size has grown significantly due to many bindings.
I have added your suggestion but if changes are not strictly related to current pull request, I would like to ask you to create a new PR next time.

msg.c
HBuczynski: The same as above.

devices/usbc/client.h
HBuczynski: Done

devices/usbc/client.c
```diff
@@ -0,0 +1,168 @@
...
+		return -1;
+	}
+
+	low_irqinst(phy_getIrq(), usbclient_intr, (void *)NULL);
+
+	while (imx_common.dc.op != DC_OP_EXIT) {
+		if (imx_common.dc.op == DC_OP_INIT) {
+			for (i = 1; i < ENDPOINTS_NUMBER; ++i) {
+				if ((res = ctrl_endptInit(i, &imx_common.data.endpts[i])) != ERR_NONE)
+					return res;
```
agkaminski, line 150: We don't need to do `usbclient_buffReset();` here? Just making sure

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+	int res = ERR_NONE;
+	u32 setup = 0;
+
+	if (endpt == 0)
+		return -1;
+
+	if ((res = ctrl_dtdInit(endpt, endpt_init->caps[USB_ENDPT_DIR_IN].init, endpt_init->caps[USB_ENDPT_DIR_OUT].init)) != ERR_NONE)
+		return res;
+
+	for(i = 0; i < ENDPOINTS_DIR_NB; ++i) {
```
agkaminski, line 151: Missing space between `for` and `(`

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+		if (endpt_init->caps[i].init)
+			setup |= ctrl_initEndptQh(endpt, i, endpt_init);
+	}
+
+	*(ctrl_common.dc->base + endptctrl0 + endpt) = setup;
+
+	for(i = 0; i < ENDPOINTS_DIR_NB; ++i) {
```
agkaminski, line 158: Missing space between `for` and `(`

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+
+
+int ctrl_endpt0Init(void)
+{
+	int i;
+	u32 qh_addr;
+
+	ctrl_allocBuff(0, USB_ENDPT_DIR_IN);
+
+	ctrl_allocBuff(0, USB_ENDPT_DIR_OUT);
```
agkaminski, line 174: In theory we might get `ERR_MEM` from them

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+
+
+int ctrl_hfIrq(void)
+{
+	int endpt = 0;
+
+	if ((ctrl_common.dc->setupstat = *(ctrl_common.dc->base + endptsetupstat)) & 0x1) {
+		/* trip winre set */
+		while (!((ctrl_common.dc->setupstat >> endpt) & 1))
+			endpt++;
```
agkaminski, line 303: Is there a chance that we end up with endpt = 32? This would cause bad memory access below.

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+		} while (!(*(ctrl_common.dc->base + usbcmd) & 1 << 13));
+
+		*(ctrl_common.dc->base + endptsetupstat) |= 1 << endpt;
+		*(ctrl_common.dc->base + usbcmd) &= ~(1 << 13);
+		*(ctrl_common.dc->base + usbsts) |= 1;
+
+		ctrl_common.dc->endptqh[0].head = ctrl_common.dc->endptqh[0].tail;
+		ctrl_common.dc->endptqh[1].head = ctrl_common.dc->endptqh[1].tail;
+
+		while (*(ctrl_common.dc->base + endptsetupstat) & 1);
```
agkaminski, line 316: `;` should be in the new line

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+
+
+int ctrl_lfIrq(void)
+{
+	if ((*(ctrl_common.dc->base + usbsts) & 1 << 6)) {
+
+		*(ctrl_common.dc->base + endptsetupstat) = *(ctrl_common.dc->base + endptsetupstat);
+		*(ctrl_common.dc->base + endptcomplete) = *(ctrl_common.dc->base + endptcomplete);
+
+		while (*(ctrl_common.dc->base + endptprime));
```
agkaminski, line 331: `;` should be in the new line

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+
+		*(ctrl_common.dc->base + endptflush) = 0xffffffff;
+
+		while (*(ctrl_common.dc->base + portsc1) & 1 << 8);
```
agkaminski, line 335: `;` should be in the new line

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,400 @@
...
+{
+	*(ctrl_common.dc->base + endptflush) = 0xffffffff;
+	/* Run/Stop bit */
+	*(ctrl_common.dc->base + usbcmd) &= ~1;
+	/* Controller resets its internal pipelines, timers etc. */
+	*(ctrl_common.dc->base + usbcmd) |= 1 << 1;
+	ctrl_common.dc->status = DC_POWERED;
+
+	/* Run/Stop register is set to 0 when the reset process is complete. */
+	while (*(ctrl_common.dc->base + usbcmd) & (1 << 1));
```
agkaminski, line 355: `;` should be in the new line

devices/usbc/desc.c
```diff
@@ -0,0 +1,349 @@
...
+	desc_common.data = usb_data_in;
+
+	for (i = 0; i < USB_BUFFER_SIZE; ++i)
+		desc_common.data->setupMem[i] =  0;
+
+	/* Extract mandatory descriptors to mapped memory */
+	for (; desList != NULL; desList = desList->next) {
+
+		if (localOffset > USB_BUFFER_SIZE)
+			return -1   ;
```
agkaminski, line 97: Extra spaces between `1` and `;`

armv7m7-imxrt106x/Makefile
```diff
@@ -9,8 +9,8 @@
 INIT_RAM = 0x0
 INIT_FLASH = 0x70000000
 
-BSS=-Tbss=20011000
-LDFLAGS+=--defsym=plo_bss=0x20011000
+BSS=-Tbss=20010000
+LDFLAGS+=--defsym=plo_bss=0x20010000
```
agkaminski, line 13: Sneaky change not really tied to the topic of the commit :) Intentional?

msg.h
```diff
@@ -37,6 +38,10 @@ typedef struct _msg_t {
 	u32 csum;
 	u32 type;
 	u8  data[MSG_MAXLEN];
+	union {
+		int (*read)(unsigned int, u8 *, u16, u16);
+		int (*write)(unsigned int, const u8 *, u16);
+	};
```
agkaminski, line 44: Should this be an union? It means that both pointers of different types point to the same function implicitly. Might cause some headache in the future

devices/usbc/ctrl.c
HBuczynski: done

devices/usbc/ctrl.c
HBuczynski: done

devices/usbc/ctrl.c
HBuczynski: Thanks, you have right.
Fixed.

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/desc.c
HBuczynski: Done

armv7m7-imxrt106x/Makefile
HBuczynski: It was intentional ;) There is allocated bigger piece of memory in bss section for USB endpoints in usbphy. Change of bss start was needed to not exceed available memory. I have tested it and there is no any influence on other system's parts.

msg.h
HBuczynski: Only one of the callback is used by` msg_t ` and it is controlled by phoenixd  so I think that union in this case is used correctly.

devices/usbc/client.c
HBuczynski: Good notice. This error means that one of the endpoints isn't initialized so that further using of driver can cause undefined behaviour.
However, I think that there should be `usbclient_destroy()` instead of `usbclient_buffReset()` because full initialization has been already set up.

devices/usbc/ctrl.c
HBuczynski: Usually host is responsible for initiating transmission with the device. So that, there is a check in line 300 whether setup transaction is received on control endpoint and then other endpoints are checked in the `while` loop.
I think that there is very little chance of that interrupts is received and no setup packet is transmitted on any of the endpoints. 

However, if we end up with`endpt > 15`, there would be definitely a bad memory access. For this reason it would be safer to check  `endpt` value before using it.

To sum up, thanks for comment ;)

devices/usbc/cdc-client.c
```diff
@@ -0,0 +1,421 @@
...
+		.bEndpointAddress = 0x80 | endpt_irq_acm0, /* direction IN */
+		.bmAttributes = 0x03,
+		.wMaxPacketSize = 0x20,
+		.bInterval = 0x08
+	},
+#if PHFS_ACM_PORTS_NB == 2
+	{
+		.bLength = 7,
+		.bDescriptorType = USB_DESC_ENDPOINT,
+		.bEndpointAddress = 0x80 | endpt_irq_acm1	, /* direction IN */
```
kemonats, line 229:
```suggestion
		.bEndpointAddress = 0x80 | endpt_irq_acm1, /* direction IN */
```

devices/usbc/client.c
```diff
@@ -0,0 +1,170 @@
...
+
+
+int usbclient_rcvEndp0(void *data, unsigned int len)
+{
+	int res = -1;
+
+	while (imx_common.dc.op != DC_OP_RCV_ENDP0)
+		;
+
+	if (imx_common.dc.op != DC_OP_RCV_ERR) {
```
kemonats, line 56: The if condition is always true.

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,410 @@
...
+{
+	u32 setup = 0;
+	int qh = endpt * 2 + dir;
+
+	ctrl_allocBuff(endpt, dir);
+
+	ctrl_common.dc->endptqh[qh].caps =  endpt_init->caps[dir].max_pkt_len << 16;
+	ctrl_common.dc->endptqh[qh].caps |= endpt_init->caps[dir].ios << 15;
+	ctrl_common.dc->endptqh[qh].caps |= endpt_init->caps[dir].zlt << 29;
+	ctrl_common.dc->endptqh[qh].caps |= endpt_init->caps[dir].mult << 30;
```
kemonats, line 129:
```suggestion
	u32 tmp;
	tmp = endpt_init->caps[dir].max_pkt_len << 16;
	tmp |= endpt_init->caps[dir].ios << 15;
	tmp |= endpt_init->caps[dir].zlt << 29;
	tmp |= endpt_init->caps[dir].mult << 30;
	ctrl_common.dc->endptqh[qh].caps = tmp;
```
What do you think about this solution.
It will be simpler (smaller size ) and faster.
`endptqh` is a pointer to `dqh_t` structure, which is with attribute `packed`.
From the GCC documentation:
```
-munaligned-access
-mno-unaligned-access

    Enables (or disables) reading and writing of 16- and 32- bit values from addresses that are not 16- or 32- bit aligned. By default unaligned access is disabled for all pre-ARMv6, all ARMv6-M and for ARMv8-M Baseline architectures, and enabled for all other architectures. If unaligned access is not enabled then words in packed data structures are accessed a byte at a time.

    The ARM attribute Tag_CPU_unaligned_access is set in the generated object file to either true or false, depending upon the setting of this option. If unaligned access is enabled then the preprocessor symbol __ARM_FEATURE_UNALIGNED is also defined.
```

devices/usbc/ctrl.c
```diff
@@ -0,0 +1,410 @@
...
+
+	ctrl_common.dc->endptqh[0].caps =  0x40 << 16; /* max 64 bytes */
+	ctrl_common.dc->endptqh[0].caps |= 0x1 << 29;
+	ctrl_common.dc->endptqh[0].caps |=  0x1 << 15; /* ios */
+	ctrl_common.dc->endptqh[0].dtd_next = 0x1; /* invalid */
+
+	ctrl_common.dc->endptqh[1].caps =  0x40 << 16;
+	ctrl_common.dc->endptqh[1].caps |= 0x1 << 29;
+	ctrl_common.dc->endptqh[1].caps |=  0x1 << 15;
+	ctrl_common.dc->endptqh[1].dtd_next = 1;
```
kemonats, line 200:
```suggestion
	u32 tmp;
	tmp =  0x40 << 16 | 0x1 << 29 | 0x1 << 15;
	ctrl_common.dc->endptqh[0].caps =  tmp;
	ctrl_common.dc->endptqh[0].dtd_next = 0x1; /* invalid */

	ctrl_common.dc->endptqh[1].caps =  tmp;
	ctrl_common.dc->endptqh[1].dtd_next = 1;
```
And the second proposal. Arguments above.
For order either `0x1` or '1'.

devices/usbc/desc.c
```diff
@@ -0,0 +1,349 @@
...
+int desc_init(usb_desc_list_t *desList, usb_common_data_t *usb_data_in, usb_dc_t *dc_in)
+{
+	int i, localOffset = 0;
+	u32 string_desc_count = 0;
+
+	desc_common.dc = dc_in;
+	desc_common.data = usb_data_in;
+
+	for (i = 0; i < USB_BUFFER_SIZE; ++i)
+		desc_common.data->setupMem[i] =  0;
```
kemonats, line 91:
```suggestion
		desc_common.data->setupMem[i] = 0;
```

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/desc.c
HBuczynski: Done

devices/usbc/ctrl.c
HBuczynski: Done

devices/usbc/client.c
HBuczynski: Done

devices/usbc/cdc-client.c
HBuczynski: Done

PR #40, opened 2021-03-25
zynq7000: add bitstream command

HBuczynski:


cmd.c
```diff
@@ -1035,3 +984,54 @@ char *cmd_getnext(char *line, unsigned int *pos, char *blanks, char *cites, char
...
+
+		return ERR_ARG;
+	}
+
+	for (i = 0; cmd_common.devs[i].name; i++)  {
+		if (!plostd_strcmp(devName, cmd_common.devs[i].name))
+			break;
+	}
+
+	if (!cmd_common.devs[i].name) {
```
agkaminski, line 1029: This `for` looks unsafe - are we certain, that we don't go outside of `devs` array? Also, we should compare pointers to NULL.

HBuczynski: A `devs` array is defined here: https://github.com/phoenix-rtos/plo/blob/ae5fedb90f4d9bc0d6ceffaa6019e29056272e60/armv7a9-zynq7000/low.c#L42-L45 so the last element in the array should be NULL.

This part of the code is derived from : https://github.com/phoenix-rtos/plo/blob/ae5fedb90f4d9bc0d6ceffaa6019e29056272e60/ia32/cmd.c#L168-L174
where objects are not compare to NULL.

In the current code, I have changed to compare pointers to NULL.

agkaminski (approved):

PR #42, opened 2021-03-30
zynq7000: add gpio controller & reset usb phy

HBuczynski:


armv7a9-zynq7000/gpio.c
```diff
@@ -0,0 +1,193 @@
...
+int gpio_writePin(u8 pin, u8 val)
+{
+	u32 v;
+	u8 bank, pos;
+
+	if (gpio_getPinPos(pin, &bank, &pos) < 0)
+		return ERR_ARG;
+
+	v = *(gpio_common.base + data + (BANK_REGS_OFFS * bank));
+	*(gpio_common.base + data + (BANK_REGS_OFFS * bank)) = (v & ~(1 << pos)) | (!!val << pos);
```
agkaminski, line 82: This is somewhat hacky - it works only thanks to the fact, that `!!` promotes `val` (`u8`) to the `int` type.

agkaminski (approved):

PR #43, opened 2021-03-30
move platforms into hal directory

HBuczynski:


agkaminski (changes requested):
Update of Makefile should be in the same commit with moving files - otherwise the code won't compile before the second commit.

agkaminski (approved):
Will squash this PR

PR #47, opened 2021-04-08
cmd: redundant calls to cmd_skipblanks() function have been removed

kemonats:
The `cmd_skipblanks()` function is called at the begining of the `cmd_getnext()` function.
I suggested optimizations of `cmd_skipblanks()` and `cmd_showprogress()` earlier in PR #15.
Btw: the `cmd_cmd()` function currently does nothing specific.

### DIFF
--- a/cmd.c
+++ b/cmd.c
@@ -297,7 +297,6 @@ void cmd_dump(char *s)
 	u8 b;
 
 	/* Get address */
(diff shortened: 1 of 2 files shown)


HBuczynski:
Your suggestions have been applied to a new version of plo.
PR is closed.

PR #48, opened 2021-04-08
Move drivers from hal into devices directory.

HBuczynski:
1. To be consistent with phoenix-rtos's naming, the serial is changed to uart.
2. Flash driver is generalized to be used by both imxrt106x and imxrt117x. **It hasn't been tested on imxrt117x.** 
3. `phfs` and `uart.h` files are moved into device directory. It is a temporary solution to keep things working. In the future, there will be introduced a new approach to initialize devices and this files will be removed.
4. `peripherals.h` is also a temporary solution. It will be replaced by dtb.


agkaminski (approved):
Looks good to me

PR #50, opened 2021-04-12
imxrt: Fixed boot from XIP leaving us unable to use ROMAPI

agkaminski:
JIRA: NIL-60


devices/flash-imxrt/flashdrv.c
```diff
@@ -280,16 +283,27 @@ int flashdrv_init(flash_context_t *ctx)
 	if ((res = flashdrv_defineFlexSPI(ctx)) < 0)
 		return res;
 
-	if (flexspi_norGetConfig(ctx->instance, &ctx->config, &ctx->option) != 0)
-		return ERR_ARG;
+	if (ctx->instance == FLASH_FLEXSPI1_INSTANCE) {
```
HBuczynski, line 286: I think that `xipStart` and `xipStop` can be defined in `flash_context_t` and then appropriate values can be written in `flashdrv_defineFlexSPI`. There will not be additional `if` statement.

PR #51, opened 2021-04-13
plo: device initialization using static constructors

HBuczynski:
This PR provides the following changes:
- add `_startc.c` to handle constructors and destructors
- add device interface to allow initialize specific driver based on platform settings

For example, registration of uart driver:

```
int uart_init(int dn, dev_handler_t *h)
{
...
}

__attribute__((constructor)) static void uart_regDev(void)
{
	devs_regDriver(DEV_UART_ID, uart_init);
}
```

If this mechanism is accepted I will implement it in the current drivers and will add changes it into this PR.


devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+ * %LICENSE%
+ */
+
+#include "devs.h"
+#include "../errors.h"
+
+#define DEV_MAX_NB            8
+#define DEV_MAX_INSTANCE_NB   16
+
+/* dev_driver_t go to .bss section where is not zeroed
```
nalajcie, line 22: "which", but the correct solution is to zero the `bss` - which can easily be added to `_start.c` function as a C code now.
This would make the `active` variable unneeded (and other istances of initializing variables to `0` not needed also).

devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+#define DRIVER_IS_ACTIVE      0xababcdcd
+
+typedef struct {
+	int (*init)(int dn, dev_handler_t *h);
+	int active;
+} dev_driver_t;
+
+
+struct {
+	dev_handler_t devs[DEV_MAX_NB][DEV_MAX_INSTANCE_NB];
```
nalajcie, line 33: not sure if we want to have two-dimension array here, seems like a waste of space (I don't know about the use cases and if it's really important)

devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+				/* TODO: check in dtb the availability of the current platform */
+				handler = &devs_common.devs[i][j];
+				driver->init(j, handler);
+			}
+		}
+	}
+}
+
+
+int devs_getHanlder(u16 dev, u16 dn, dev_handler_t **h)
```
nalajcie, line 77: `devs_getHanlder` - typo

devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+			handler->read = NULL;
+			handler->write = NULL;
+		}
+
+		driver = &devs_common.drivers[i];
+		if (driver->active == DRIVER_IS_ACTIVE) {
+			for (j = 0; j < DEV_MAX_INSTANCE_NB; ++j) {
+				/* TODO: check in dtb the availability of the current platform */
+				handler = &devs_common.devs[i][j];
+				driver->init(j, handler);
```
nalajcie, line 70: I assume this will be able to fail in the future and report it by return code?

devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+} dev_driver_t;
+
+
+struct {
+	dev_handler_t devs[DEV_MAX_NB][DEV_MAX_INSTANCE_NB];
+	dev_driver_t drivers[DEV_MAX_NB];
+} devs_common;
+
+
+void devs_regDriver(int dev, int (*init)(int dn, dev_handler_t *h))
```
nalajcie, line 38: `int dev` should probably be `unsigned int dev`, currently I can call `devs_regDriver(-3, NULL)` and overwrite BSS before the array :)

probably `int dn` should also be `unsigned`

devices/devs.c
```diff
@@ -0,0 +1,153 @@
...
+
+
+void devs_init(void)
+{
+	u32 i, j;
+	dev_driver_t *driver;
+	dev_handler_t *handler;
+
+	for (i = 0; i < DEV_MAX_NB; ++i) {
+		for (j = 0; j < DEV_MAX_INSTANCE_NB; ++j) {
```
nalajcie, line 58: would not be needed when we zero the BSS

devices/devs.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * %LICENSE%
+ */
+
+
+#ifndef _DEVS_H_
+#define _DEVS_H_
+
+#include "../types.h"
+#include "config.h"
```
nalajcie, line 21: is `config.h` needed for this header to parse?

devices/devs.h
```diff
@@ -0,0 +1,61 @@
...
+
+typedef struct {
+	int (*deinit)(u16 dn);
+
+	ssize_t (*read)(u16 dn, addr_t offs, u8 *buff, u32 len);
+	ssize_t (*write)(u16 dn, addr_t offs, const u8 *buff, u32 len);
+} dev_handler_t;
+
+
+extern void devs_regDriver(int dev, int (*init)(int dn, dev_handler_t *h));
```
nalajcie, line 41: I would add comment that this function can be called only before `devs_init`, preferably from device drivers constructors

nalajcie (commented):
Some things need polishing, but I think it's the move in the right direction

devices/devs.c
HBuczynski: Fixed

devices/devs.c
HBuczynski: Yes, you have right. An oversight on my part.

devices/devs.h
HBuczynski: `config.h` contains platform specific typedefs. In this case, the `ssize_t` is located in config header.

devices/devs.h
HBuczynski: Ok, it can be helpful.

devices/devs.c
HBuczynski: The report error will be printed by debug console which will not be depend on uart driver. 
I also wondering about plo behaviour when one of the driver will not be initialized correctly. Should it only report error or terminate as well?

devices/devs.c
HBuczynski: The two-dimension array was introduced here to keep handlers of instances of all devices. For example, imxrt107x has 12 UART instances so that we need to be ready to keep them all.

devices/devs.c
nalajcie: In production environment I would try to boot the device nevertheless (who cares about uart5 anyway? :p). For debugging I think You will probably want to stop early to debug, but that can be done by placing asserts (either here or preferably at the point of failure - so in the device init function)

devices/devs.c
nalajcie: yeah, I understand WHY we need it (it's major/minor), just usually we will have 12 uarts and single ethernet / flash/ etc, so for most devices the other instances would be empty. I like the simplicity so If we're not afraid of the memory usage it's ok. We can always change the implementation to hashtable/linked list if that would provide to be problem on smaller targets.

_startc.c
```diff
@@ -0,0 +1,43 @@
...
+
+extern void plo_init(void);
+
+
+void _startc(int argc, char **argv, char **env)
+{
+	size_t i, size;
+
+	/* Clear the .bss section */
+	hal_memset(plo_bss, 0, (addr_t)_end - (addr_t)plo_bss);
```
kemonats, line 32: `_plo_bss` not  defined in `TARGET=armv7a9-zynq7000`

hal/armv7a/zynq7000/hal.c
```diff
@@ -190,6 +190,35 @@ void hal_memcpy(void *dst, const void *src, unsigned int l)
 }
 
 
+void hal_memset(void *dst, int v, unsigned int l)
```
kemonats, line 193: If you are interested  I can provide optimized version of the `hal_memset()` function ( smaler size and uses fewer registers ).

HBuczynski: This code comes from [phoenix-rtos-kernel/hal/armv7a](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/hal/armv7a/string.h).
If you would like to optimize it, go ahead. However, I would like to ask you to do it in separate PR associated with phoenix-rtos-kernel. This thread hones in on devs module so I wouldn't like to discuss others, not related topics here.

kemonats: I proposed this almost a year ago PR [#88](https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/88)

_startc.c
HBuczynski: Fixed

devices/devs.c
HBuczynski: Ok

devices/devs.c
HBuczynski: This initialization is already removed, BSS is zeroed.

devices/devs.c
HBuczynski: BSS is zeroed. `active` variable has been removed.

hal/armv7a/zynq7000/hal.c
HBuczynski: I will test it and add in a separate PR.

hal/armv7a/zynq7000/Makefile
```diff
@@ -11,6 +11,7 @@ INIT_FLASH = 0x0
 
 LDFLAGS:=$(filter-out -Tbss% , $(LDFLAGS))
 LDFLAGS:=$(filter-out -Tdata% , $(LDFLAGS))
+LDFLAGS+=--defsym=_plo_bss=__bss_start
```
nalajcie, line 14: ok, now not so sure it will work, as the value of `__bss_start` will be `0`

hal/armv7a/zynq7000/config.h
```diff
@@ -40,7 +40,7 @@
 
 /* Linker symbols */
 extern void _end(void);
-extern void plo_bss(void);
+extern void _plo_bss(void);
```
nalajcie, line 43: maybe fix the linker symbols declaration here also?

nalajcie (changes requested):
I think You have linker variable addresses and values mixed up

_startc.c
```diff
@@ -0,0 +1,45 @@
...
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "hal.h"
+
+extern void _end(void);
```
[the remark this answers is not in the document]
agkaminski: This symbol is just a symbol with no type. We have to give it some type to declare it in C. Declaring it as a function has an advantage that we can access this sybol without `&`. To get _end we need either:

```
extern char _end;

address = &_end;
```

or 
```

extern void _end(void);

address = _end;
```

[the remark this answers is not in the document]
nalajcie: I wrote the same in my next comment :). I believe directly accessing the variable address is more readable than defining it as a function and using the name as a function pointer (seems to me that the "more advanced knowledge of C" is needed to understand the function pointers syntax - for casual reader it seems we're using the `_end` value, not it's address).

The main issue with `_plo_bss` still exists (we're setting symbol **value** - not **address** - by `LDFLAGS` - and here by declaring it as a function - we use it's **address**). Actually, the existence of this bug is a proof that the previously used syntax is somewhat hard :P.

I'm fine with either approach, it just comes down to the readability.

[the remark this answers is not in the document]
nalajcie: sorry, my experiments on host target were invalid, indeed by using `--defsym` `LDFLAG` You would directly set the symbol **address**, the symbol will not be put in any section (will be marked as absent).

So the code works correctly (the only consideration is readability).

_startc.c
```diff
@@ -0,0 +1,45 @@
...
+
+extern void plo_init(void);
+
+
+void _startc(int argc, char **argv, char **env)
+{
+	size_t i, size;
+
+	/* Clear the .bss section */
+	hal_memset(_plo_bss, 0, (addr_t)_end - (addr_t)_plo_bss);
```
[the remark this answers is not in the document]
nalajcie: sorry, my experiments on host target were invalid, indeed by using --defsym LDFLAG You would directly set the symbol address, the symbol will not be put in any section (will be marked as absent).

So the code works correctly (the only consideration is readability).
Closing this as the code should work.

devices/devs.c
```diff
@@ -0,0 +1,132 @@
...
+	if (dev >= SIZE_DEV || dn >= SIZE_DEV_INSTANCES)
+		return -ERR_ARG;
+
+	temph = &devs_common.devs[dev][dn];
+	if (temph->deinit == NULL ||
+		temph->read == NULL ||
+		temph->write == NULL)
+		return -ERR_ARG;
+
+	*h = &devs_common.devs[dev][dn];
```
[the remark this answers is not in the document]
HBuczynski: Done

nalajcie (approved):
LGTM

cmd.c
```diff
@@ -897,6 +882,75 @@ void cmd_syspage(char *s)
 }
 
 
+void cmd_phfs(char *s)
+{
+	u16 argsc = 0;
+	unsigned int i, major, minor, pos = 0;
+	char args[MAX_CMD_ARGS_NB][LINESZ + 1];
```
kemonats, line 889: Since the maximum number of arguments is 4, it can declare 
```
	char args[5][LINESZ + 1]
```
save `5 * 81 ` bytes on the stack.

cmd.c
```diff
@@ -897,6 +882,75 @@ void cmd_syspage(char *s)
...
+
+	for (i = 0; argsc < MAX_CMD_ARGS_NB; ++i) {
+		cmd_skipblanks(s, &pos, ". \t");
```
kemonats, line 892: The `cmd_skipblanks()` function is called at the begining of the `cmd_getnext()` function.

cmd.c
```diff
@@ -897,6 +882,75 @@ void cmd_syspage(char *s)
...
+}
+
+
+void cmd_devs(char *s)
+{
+	u16 argsc = 0;
+	char args[MAX_CMD_ARGS_NB][LINESZ + 1];
+
+	/* Parse command arguments */
+	if (cmd_parseArgs(s, args, &argsc) > 0) {
```
kemonats, line 918: If it will not been expanded:
```suggestion
	unsigned int pos = 0;

	cmd_skipblanks(s, &p, DEFAULT_BLANKS);
	s += p;

	if (*s) {
```

cmd.c
```diff
@@ -897,6 +882,75 @@ void cmd_syspage(char *s)
...
+
+
+void cmd_console(char *s)
+{
+	u16 argsc = 0;
+	unsigned int i, major, minor, pos = 0;
+	char args[MAX_CMD_ARGS_NB][LINESZ + 1];
+
+	for (i = 0; argsc < MAX_CMD_ARGS_NB; ++i) {
+		cmd_skipblanks(s, &pos, ". \t");
```
kemonats, line 934: The `cmd_skipblanks()` function is called at the begining of the `cmd_getnext()` function.

phfs.c
```diff
@@ -13,61 +13,362 @@
...
+{
+	int i;
+	size_t sz;
+	phfs_handler_t *ph;
 
-	return phfs_handlers[pdn].open(phfs_handlers[pdn].dn, name, flags);
+	for (i = 0; i < phfs_common.hCnt; ++i) {
+		ph = &phfs_common.handlers[i];
+		sz = plostd_strlen(ph->alias);
+		if ((sz == plostd_strlen(alias)) && (plostd_strncmp(alias, ph->alias, sz) == 0))
```
kemonats, line 98: Move `plostd_strlen(alias)` out of `for` loop, `alias` length does not change in a loop.

phfs.c
```diff
@@ -13,61 +13,362 @@
...
+static int phfs_getFileId(const char *alias)
+{
+	int i;
+	size_t sz;
+	phfs_file_t *file;
+
+	for (i = 0; i < phfs_common.fCnt; ++i) {
+		file = &phfs_common.files[i];
+		sz = plostd_strlen(file->alias);
+		if ((sz == plostd_strlen(alias)) && (plostd_strncmp(alias, file->alias, sz) == 0))
```
kemonats, line 178: Same as in `phfs_getHandlerId(const char *alias)`

phfs.c
```diff
@@ -13,61 +13,362 @@
...
+	/* Check if alias is already in use */
+	if (phfs_getFileId(alias) >= 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - alias is already in use", alias);
 		return ERR_ARG;
+	}
+
+	file = &phfs_common.files[phfs_common.fCnt];
 
-	return phfs_handlers[pdn].write(phfs_handlers[pdn].dn, handle, pos, buff, len, sync);
+	sz = (plostd_strlen(alias) < 31) ? (plostd_strlen(alias) + 1) : 31;
```
kemonats, line 207: same as in `int phfs_regDev(const char *alias, unsigned int major, unsigned int minor, const char *prot)`

cmd.c
HBuczynski: You have right. Fixed

cmd.c
HBuczynski: You have right. Fixed

cmd.c
HBuczynski: Done

cmd.c
```diff
@@ -897,6 +882,75 @@ void cmd_syspage(char *s)
...
+
+	phfs_showDevs();
+}
+
+
+void cmd_console(char *s)
+{
+	u16 argsc = 0;
+	unsigned int i, major, minor, pos = 0;
+	char args[MAX_CMD_ARGS_NB][LINESZ + 1];
```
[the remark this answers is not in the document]
HBuczynski: You have right. Fixed

cmd.c
HBuczynski: Removed `cmd_skipblanks()`

phfs.c
HBuczynski: Thanks for noticing. Fixed

phfs.c
HBuczynski: Thanks for noticing. Fixed

phfs.c
HBuczynski: Done

phfs.c
```diff
@@ -13,61 +13,362 @@
...
+	}
+
+	ph = &phfs_common.handlers[phfs_common.hCnt];
+	if (phfs_setProt(ph, prot) < 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - wrong protocol name\n\t use: \"%s\", \"%s\"", prot,
+		              phfs_getProtName(phfs_prot_raw), phfs_getProtName(phfs_prot_phoenixd));
+		return ERR_ARG;
+	}
+
+	sz = (plostd_strlen(alias) < 7) ? (plostd_strlen(alias) + 1) : 7;
```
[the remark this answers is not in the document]
HBuczynski: Ok, I agree with you, code is more readable and `plostd_strlen` is invoked only once.

nalajcie:
generic comment regarding the development process:
 - I don't like that after I approved the PR there were additional commits added
 - let's not treat PRs as a long-lasting feature branch - let's just create the branch and make small approved PRs against it - and then rebase/merge the feature branch to master once it's ready
 - making PR grow in size makes it muuuch harder to review
 - (side note: if You want to mark PR as a draft - there is a dedicated button for that, don't use "DRAFT" label)
 - "draft" PRs mean they are not ready for review and I think nobody should review them. We already reviewed this PR so it should not be marked as draft.

Because of the above I'm withdrawing my PR approval and don't plan to review this PR anymore.

nalajcie (commented):
PR approval removed.

HBuczynski:
Due to the fact that this PR has grown to big size, it will be divided into the smaller ones to make changes more readable during review.
Comments discussed in this section are moved to other PRs.

PR #53, opened 2021-04-19
script: script parsing improved

kemonats:
Instead of creating an array of copies of individual script lines, I create an array of pointers to those lines.

Other changes:
- I moved the calculation of name length and parameter position outside the for loop.
- I changed the argument of the `script_expandAlias()` function from `char **name` to `char *name` (in the `script_expandAlias()` function, the pointer address to the name does not change).

The size of the `script_common` variable:
```
sizeof(script_common)
now           new
4+64*80=5124  4+64*4=260
```

What do you think about such a solution:
`script` variable as:
```
static char *script[] = {
    #include <script.plo.h>
    NULL
};
```
and the file `script.plo.h` is created in the `Makefile`.
I can show such a solution as PR.

### DIFF
--- a/cmd.c
+++ b/cmd.c
@@ -737,7 +737,7 @@ void cmd_app(char *s)
 
 	pos = plostd_strlen(name);
 	if (name[0] == '@') {
-		if (script_expandAlias(&name) < 0) {
+		if (script_expandAlias(name) < 0) {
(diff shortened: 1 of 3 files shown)


kemonats:
This code will not work because the script variable is stored in flash memory.

PR #54, opened 2021-04-22
plo: constructors, destructors and device module

HBuczynski:
Due to the fact that [PR](https://github.com/phoenix-rtos/plo/pull/51) has been growing, it is separated into different PRs which will be merged into a `feature/devs` branch in plo. When the whole functionality will done, the `feature/devs` will be merged into master.

This PR provides the following changes:
- add `_startc.c` to handle constructors and destructors
- add device interface to allow initialize specific driver based on platform settings

For example, registration of uart driver:

```
int uart_init(int dn, dev_handler_t *h)
{
...
}

__attribute__((constructor)) static void uart_regDev(void)
{
	devs_regDriver(DEV_UART_ID, uart_init);
}
```

If this mechanism is accepted I will implement it in the current drivers and will add changes it into this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ include hal/$(TARGET_SUFF)/$(TARGET_SUBFAMILY)/Makefile
 CFLAGS += $(BOARD_CONFIG)
 CFLAGS += -I../plo/hal/$(TARGET_SUFF)/$(TARGET_SUBFAMILY)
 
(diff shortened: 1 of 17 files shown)


nalajcie (approved):
assuming these are the 2 original commits already reviewed, LGTM to feature branch.

PR #55, opened 2021-04-22
phfs: add new interface

HBuczynski:
Some of the commands were disable. They will be updated in the other PR.

### DIFF
--- a/phfs.c
+++ b/phfs.c
@@ -13,61 +13,383 @@
  * %LICENSE%
  */
 
-#include "errors.h"
-#include "hal.h"
-#include "plostd.h"
 #include "phfs.h"
+#include "devs.h"
+#include "hal.h"
 
-/* Configuration file for specific platform */
-#include "config.h"
+#include "phoenixd.h"
+#include "plostd.h"
+#include "errors.h"
 
 
-phfs_handler_t phfs_handlers[PDN_NB];
+#define SIZE_PHFS_HANDLERS    8
+#define SIZE_PHFS_FILES       10
 
 
-handle_t phfs_open(u16 pdn, const char *name, u32 flags)
+
+enum { phfs_prot_raw = 0, phfs_prot_phoenixd };
+
+
+typedef struct {
+	char alias[8];
+	unsigned int major;
+	unsigned int minor;
+	unsigned int prot;
+
+	dev_handler_t *h;
+} phfs_handler_t;
+
+
+typedef struct {
+	char alias[32];
+	addr_t addr;
+	size_t size;
+} phfs_file_t;
+
+
+struct {
+	phfs_handler_t handlers[SIZE_PHFS_HANDLERS];
+	unsigned int hCnt;
+
+	phfs_file_t files[SIZE_PHFS_FILES];
+	unsigned int fCnt;
+} phfs_common;
+
+
+static const char *phfs_getProtName(unsigned int id)
 {
-	handle_t handle;
+	switch (id) {
+	case phfs_prot_raw:
+		return "raw";
 
-	if (pdn >= PDN_NB) {
-		handle.h = ERR_ARG;
-		return handle;
+	case phfs_prot_phoenixd:
+		return "phoenixd";
+
+	default:
+		return "";
+	}
+}
+
+
+static int phfs_setProt(phfs_handler_t *ph, const char *prot)
+{
+	if (prot == NULL || (plostd_strcmp(prot, "raw") == 0))
+		ph->prot = phfs_prot_raw;
+	else if (plostd_strcmp(prot, "phoenixd") == 0)
+		ph->prot = phfs_prot_phoenixd;
+	else
+		return ERR_ARG;
+
+	return ERR_NONE;
+}
+
+
+static int phfs_getHandlerId(const char *alias)
+{
+	int i;
+	phfs_handler_t *ph;
+
+	for (i = 0; i < phfs_common.hCnt; ++i) {
+		ph = &phfs_common.handlers[i];
+		if ((plostd_strcmp(alias, ph->alias) == 0))
+			return i;
 	}
 
-	return phfs_handlers[pdn].open(phfs_handlers[pdn].dn, name, flags);
+	return ERR_ARG;
 }
 
 
-s32 phfs_read(u16 pdn, handle_t handle, addr_t *pos, u8 *buff, u32 len)
+int phfs_regDev(const char *alias, unsigned int major, unsigned int minor, const char *prot)
 {
-	if (pdn >= PDN_NB)
+	size_t sz;
+	dev_handler_t *dh;
+	phfs_handler_t *ph;
+
+	if (alias == NULL)
 		return ERR_ARG;
 
-	return phfs_handlers[pdn].read(phfs_handlers[pdn].dn, handle, pos, buff, len);
+	if (phfs_common.hCnt >= SIZE_PHFS_HANDLERS) {
+		plostd_printf(ATTR_ERROR, "\nToo many devices");
+		return ERR_ARG;
+	}
+
+	if (devs_getHandler(major, minor, &dh) < 0) {
+		plostd_printf(ATTR_ERROR, "\n%d.%d - device does not exist", major, minor);
+		return ERR_ARG;
+	}
+
+	/* Check if alias is already in use */
+	if (phfs_getHandlerId(alias) >= 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - alias is already in use", alias);
+		return ERR_ARG;
+	}
+
+	ph = &phfs_common.handlers[phfs_common.hCnt];
+	if (phfs_setProt(ph, prot) < 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - wrong protocol name\n\t use: \"%s\", \"%s\"", prot,
+		              phfs_getProtName(phfs_prot_raw), phfs_getProtName(phfs_prot_phoenixd));
+		return ERR_ARG;
+	}
+
+	sz = plostd_strlen(alias);
+	sz = (sz < (sizeof(ph->alias) - 1)) ? (sz + 1) : (sizeof(ph->alias) - 1);
+	hal_memcpy(ph->alias, alias, sz);
+	ph->alias[sz] = '\0';
+
+	ph->h = dh;
+	ph->major = major;
+	ph->minor = minor;
+	phfs_common.hCnt++;
+
+	return ERR_NONE;
 }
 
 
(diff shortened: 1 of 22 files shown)


phfs.c
```diff
@@ -13,61 +13,393 @@
...
+	phfs_handler_t *ph;
+
+	if ((res = phfs_getHandlerId(alias)) < 0)
+		return ERR_ARG;
+
+	handler->pn = res;
+	ph = &phfs_common.handlers[handler->pn];
+
+	switch (ph->prot)
+	{
```
agkaminski, line 257: `switch (ph->prot) {`

phfs.c
```diff
@@ -13,61 +13,393 @@
...
+	phfs_file_t *file;
+	phfs_handler_t *ph;
+
+	if (handler.pn >= SIZE_PHFS_HANDLERS)
+		return ERR_ARG;
+
+	ph = &phfs_common.handlers[handler.pn];
+
+	switch (ph->prot)
+	{
```
agkaminski, line 300: `switch (ph->prot) {`

phfs.c
```diff
@@ -13,61 +13,393 @@
...
+	phfs_handler_t *ph;
+	phfs_file_t *file;
+
+	if (handler.pn >= SIZE_PHFS_HANDLERS)
+		return ERR_ARG;
+
+	ph = &phfs_common.handlers[handler.pn];
+
+	switch (ph->prot)
+	{
```
agkaminski, line 343: `switch (ph->prot) {`

phfs.c
```diff
@@ -13,61 +13,393 @@
...
+	phfs_clbk_t clbk;
+	phfs_handler_t *ph;
+
+	if (handler.pn >= SIZE_PHFS_HANDLERS)
+		return ERR_ARG;
+
+	ph = &phfs_common.handlers[handler.pn];
+
+	switch (ph->prot)
+	{
```
agkaminski, line 386: `switch (ph->prot) {`

phfs.h
```diff
@@ -19,29 +21,54 @@
 #include "config.h"
 
 typedef struct {
-	s32 h;
-	void *offs;
-} handle_t;
+	unsigned int pn; /* phfs device number descriptor */
+	unsigned int fd; /* file descriptor               */
```
agkaminski, line 25: It think we agreeg on "file handle" nomenclature

agkaminski (commented):
Typo in "hal: remove the remainf of the old phfs and devs " commit name

phfs.c
HBuczynski: Done

phfs.c
HBuczynski: Done

phfs.c
HBuczynski: Done

phfs.c
HBuczynski: Done

phfs.h
HBuczynski: I remember that we should use "device handle" instead of "device descriptor" so probably I should change comment `/* phfs device number descriptor */` to `/* phfs device handle */`.
However, we didn't have discussion about files. I assumed that **file descriptor** is a common phrase and it can be applied in this case.

phfs.c
```diff
@@ -13,61 +13,388 @@
...
 
-	return phfs_handlers[pdn].open(phfs_handlers[pdn].dn, name, flags);
+static int phfs_setProt(phfs_handler_t *ph, const char *prot)
+{
+	size_t sz = plostd_strlen(prot);
+
+	if (prot == NULL || (sz == 3 && plostd_strncmp(prot, "raw", 3) == 0))
+		ph->prot = phfs_prot_raw;
+	else if (sz == 8 && plostd_strncmp(prot, "phoenixd", 8) == 0)
+		ph->prot = phfs_prot_phoenixd;
```
kemonats, line 80:
```suggestion
	if (prot == NULL || (plostd_strcmp(prot, "raw") == 0))
		ph->prot = phfs_prot_raw;
	else if (plostd_strcmp(prot, "phoenixd") == 0)
		ph->prot = phfs_prot_phoenixd;
```
`plostd_strcmo()` compares strings as to value and length.

phfs.c
```diff
@@ -13,61 +13,388 @@
...
+	int i;
+	phfs_handler_t *ph;
+	size_t sz, aliasSz = plostd_strlen(alias);
+
+	for (i = 0; i < phfs_common.hCnt; ++i) {
+		ph = &phfs_common.handlers[i];
+		sz = plostd_strlen(ph->alias);
+		if ((sz == aliasSz) && (plostd_strncmp(alias, ph->alias, sz) == 0))
+			return i;
+	}
```
kemonats, line 99:
```suggestion

	for (i = 0; i < phfs_common.hCnt; ++i) {
		ph = &phfs_common.handlers[i];
		if (plostd_strcmp(alias, ph->alias) == 0)
			return i;
	}
```
Same as `phfs_setProt()`.

phfs.c
```diff
@@ -13,61 +13,388 @@
...
+
+	ph = &phfs_common.handlers[phfs_common.hCnt];
+	if (phfs_setProt(ph, prot) < 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - wrong protocol name\n\t use: \"%s\", \"%s\"", prot,
+		              phfs_getProtName(phfs_prot_raw), phfs_getProtName(phfs_prot_phoenixd));
+		return ERR_ARG;
+	}
+
+	sz = plostd_strlen(alias);
+	sz = (sz < 7) ? (sz + 1) : 7;
```
kemonats, line 138: Instead of magic number `7` use `sizeof(ph->alias) - 1`.
Maybe add `plostd_strncpy()` to such copies?

phfs.c
```diff
@@ -13,61 +13,388 @@
...
+	int i;
+	phfs_file_t *file;
+	size_t sz, aliasSz = plostd_strlen(alias);
+
+	for (i = 0; i < phfs_common.fCnt; ++i) {
+		file = &phfs_common.files[i];
+		sz = plostd_strlen(file->alias);
+		if ((sz == aliasSz) && (plostd_strncmp(alias, file->alias, sz) == 0))
+			return i;
+	}
```
kemonats, line 202:
```suggestion

	for (i = 0; i < phfs_common.fCnt; ++i) {
		file = &phfs_common.files[i];
		if (plostd_strcmp(alias, file->alias) == 0)
			return i;
	}
```
Same as `phfs_setProt()`.

phfs.c
```diff
@@ -13,61 +13,388 @@
...
+	/* Check if alias is already in use */
+	if (phfs_getFileId(alias) >= 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - alias is already in use", alias);
+		return ERR_ARG;
+	}
+
+	file = &phfs_common.files[phfs_common.fCnt];
+
+	sz = plostd_strlen(alias);
+	sz = (sz < 31) ? (sz + 1) : 31;
```
kemonats, line 230: Instead of magic number `31` use `sizeof(file->alias) - 1`.

script.c
```diff
@@ -16,78 +16,90 @@
...
+			addr = plostd_ahtoi(alias[i]);
+		}
+		else if (i == 2) {
+			if (plostd_ishex(alias[i]) < 0) {
+				return ERR_ARG;
+			}
+
+			sz = plostd_ahtoi(alias[i]);
+
+		}
```
kemonats, line 60: This can be simplified to something like
```suggestion
		if (i == 0)
			continue;

		if (plostd_ishex(alias[i]) < 0){
			return ERR_ARG;

		if (i == 1) {
			addr = plostd_ahtoi(alias[i]);
		}
		else if (i == 2) {
			sz = plostd_ahtoi(alias[i]);
		}
```

script.c
```diff
@@ -16,78 +16,90 @@
...
+}
+
+
 void script_init(void)
 {
 	int i;
 	unsigned int pos = 0;
 	char *argsScript = (char *)script;
 
-	script_common.cmdCnt = 0;
```
kemonats, line 39: Why are you removing this line?

phfs.c
HBuczynski: Yes, you have right. Changed to `plostd_strcmp()`

phfs.c
HBuczynski: Yes, you have right. Changed to `plostd_strcmp()`

phfs.c
HBuczynski: Yes, you have right. Changed to `plostd_strcmp()`

script.c
HBuczynski: Done

script.c
HBuczynski: In this [**PR**](https://github.com/phoenix-rtos/plo/pull/54) the `.bss` section is zeroed so that this variable does not have to be initialized again in this function if we assume that `script_init()` is invoked only once.

phfs.c
HBuczynski: Yes, we can add `plostd_strncpy()` but not in this PR.

phfs.c
HBuczynski: Done

agkaminski (approved):
Generally LGTM. There is still an open matter of nomenclature, but it can be solved in a future PR.

phfs.c
```diff
@@ -13,61 +13,383 @@
...
+	if (phfs_setProt(ph, prot) < 0) {
+		plostd_printf(ATTR_ERROR, "\n%s - wrong protocol name\n\t use: \"%s\", \"%s\"", prot,
+		              phfs_getProtName(phfs_prot_raw), phfs_getProtName(phfs_prot_phoenixd));
+		return ERR_ARG;
+	}
+
+	sz = plostd_strlen(alias);
+	sz = (sz < (sizeof(ph->alias) - 1)) ? (sz + 1) : (sizeof(ph->alias) - 1);
+	hal_memcpy(ph->alias, alias, sz);
+	ph->alias[sz] = '\0';
```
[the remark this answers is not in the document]
HBuczynski: Currently, we don't have implementation of `strncpy`. We are going to  make modification in `plostd` module so it will be also a good time to add some extra functions.

phfs.h
pawelpisarczyk: We should change it to the handle. Semantically file descriptor plays role of resource handle which points to the operating system resource shared among running processes. In this case we have no processes but to be inline with our agreements and  be consistent I suggest not to introduce new terms.

phfs.c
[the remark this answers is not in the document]
pawelpisarczyk: We have no implementation of strncpy and we don't plan to link libphoenix or internal compiler primitives, so I think this approach is correct.

PR #56, opened 2021-04-22
cmds: update to new interface of phfs

HBuczynski:
This PR should be started reviewing when [this one](https://github.com/phoenix-rtos/plo/pull/55) will be merged and `cmds` branch will be rebased.

### DIFF
--- a/cmd.c
+++ b/cmd.c
@@ -377,9 +377,14 @@ void cmd_write(char *s)
 }
 
 
-static int cmd_parseDev(handler_t *h, addr_t *offs, size_t *sz, const char (*args)[LINESZ + 1], u16 *argsID, u16 argsc)
+static int cmd_parseDev(handler_t *h, addr_t *offs, size_t *sz, char (*args)[LINESZ + 1], u16 *argsID, u16 argsc)
 {
-	const char *alias = args[(*argsID)++];
+	char *alias;
+
+	if (h == NULL || offs == NULL || sz == NULL || args == NULL || argsID == NULL)
+		return ERR_ARG;
+
+	alias = args[(*argsID)++];
 
 	if ((*argsID) >= argsc) {
 		plostd_printf(ATTR_ERROR, "\nWrong number of arguments\n");
@@ -432,13 +437,12 @@ void cmd_copy(char *s)
 	}
 
 	if (cmd_parseDev(&h[0], &offs[0], &sz[0], args, &argsID, argsc) < 0) {
-		phfs_close(h[0]);
+		plostd_printf(ATTR_ERROR, "\nCannot open file\n");
 		return;
 	}
 
 	if (cmd_parseDev(&h[1], &offs[1], &sz[1], args, &argsID, argsc) < 0) {
-		phfs_close(h[0]);
-		phfs_close(h[1]);
+		plostd_printf(ATTR_ERROR, "\nCannot open file\n");
 		return;
 	}
 
@@ -691,32 +695,20 @@ void cmd_app(char *s)
 		if (cmd_getnext(cmdArgs[argID], &pos, "@ (:) \t", NULL, appData[i], sizeof(appData[i])) == NULL || *appData[i] == 0)
 				break;
 
-		switch (i) {
-			case 0:
-				cmdline = appData[0];
-				break;
-
-			case 1:
-				if (plostd_ishex(appData[i]) < 0) {
-					plostd_printf(ATTR_ERROR, "\nOffset is not a hex value !!\n");
-					return;
-				}
-
-				offs = plostd_ahtoi(appData[i]);
-				break;
-
-			case 2:
-				if (plostd_ishex(appData[i]) < 0) {
-					plostd_printf(ATTR_ERROR, "\nSize is not a hex value !!\n");
-					return;
-				}
-
-				sz = plostd_ahtoi(appData[i]);
-				break;
-
-			default:
-				break;
+		if (i == 0) {
+			cmdline = appData[0];
+			continue;
 		}
+
+		if (plostd_ishex(appData[i]) < 0) {
+			plostd_printf(ATTR_ERROR, "\nOffset is not a hex value !!\n");
+			return;
+		}
+
+		if (i == 1)
+			offs = plostd_ahtoi(appData[i]);
+		else if (i == 2)
+			sz = plostd_ahtoi(appData[i]);
 	}
 
 	/* Get app name from cmdline */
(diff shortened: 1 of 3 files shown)


cmd.c
```diff
@@ -384,89 +375,81 @@ void cmd_write(char *s)
...
 	}
 
-	/* Parse source parameters */
-	if (cmd_checkDev(cmdArgs[argsID], &phfses[0].dn) < 0)
-		return;
-
-	if (cmd_parseDev(&phfses[0], cmdArgs, &argsID, cmdArgsc) < 0) {
-		phfs_close(phfses[0].dn, phfses[0].handle);
+	if (cmd_parseDev(&h[0], &offs[0], &sz[0], args, &argsID, argsc) < 0) {
+		phfs_close(h[0]);
```
agkaminski, line 435: If we failed to parseDev do we need to close the handler? I would assume that it shouldn't have been opened in that case, so there is nothing to close.

cmd.c
```diff
@@ -843,6 +843,75 @@ void cmd_syspage(char *s)
...
+	for (i = 0; argsc < 5; ++i) {
+		if (cmd_getnext(s, &pos, ". \t", NULL, args[i], sizeof(args[i])) == NULL || *args[i] == 0)
+			break;
+		argsc++;
+	}
+
+	if (argsc < 3 || argsc > 4) {
+		plostd_printf(ATTR_ERROR, "\nWrong arguments!!\n");
+		return;
+	}
```
agkaminski, line 861: First we parse arguments, then we check how many there were? I think that `if` and `for` should switch places.

cmd.c
```diff
@@ -843,6 +843,75 @@ void cmd_syspage(char *s)
...
+	unsigned int pos = 0;
+
+	cmd_skipblanks(s, &pos, DEFAULT_BLANKS);
+	s += pos;
+
+	if (*s) {
+		plostd_printf(ATTR_ERROR, "\nCommand should not consist of any args\n");
+		return;
+	}
+
```
agkaminski, line 883: "Command foobar does not take any arguments" perhaps?

We might ignore any additional arguments, shouldn't do us any harm.

cmd.c
```diff
@@ -843,6 +843,75 @@ void cmd_syspage(char *s)
...
+	char args[3][LINESZ + 1];
+
+	for (i = 0; argsc < 3; ++i) {
+		if (cmd_getnext(s, &pos, ". \t", NULL, args[i], sizeof(args[i])) == NULL || *args[i] == 0)
+			break;
+		argsc++;
+	}
+
+	if (argsc != 2) {
+		plostd_printf(ATTR_ERROR, "\nWrong arguments!!\n");
```
agkaminski, line 901: "Wrong number of arguments" perhaps?

hal/armv7a/zynq7000/cmd.c
```diff
@@ -70,5 +64,4 @@ void cmd_bitstream(char *s)
 	}
 
 	plostd_printf(ATTR_LOADER, "PL was successfully initialized\n");
-#endif
-}
+}
```
agkaminski, line 67: No new line at the end of the file

agkaminski (changes requested):

cmd.c
HBuczynski: As you can see in `for` loop there is incrementation of `argsc`. We don't  know how many arguments is in the command so firstly you have to parse them and after that check the count.

cmd.c
HBuczynski: I assume that passing wrong arguments is an incorrect usage of command.

cmd.c
agkaminski: Yeah, you're right, I missed that. There is no real need for `i` in this loop. How about:

```
for (argsc  = 0; argsc < 5; ++argsc ) {
	if (cmd_getnext(s, &pos, ". \t", NULL, args[argsc], sizeof(args[argsc])) == NULL || *args[argsc] == '\0')
		break;
}
```

cmd.c
agkaminski: Ok, I have no problem with that ;)

cmd.c
```diff
@@ -724,86 +703,77 @@ void cmd_app(char *s)
...
-			phfs.dataOffs = plostd_ahtoi(appData[i]);
+			default:
+				break;
 		}
-		else if (i == 2) {
-			if (plostd_ishex(appData[i]) < 0) {
-				plostd_printf(ATTR_ERROR, "\nSize is not a hex value !!\n");
-				return;
-			}
+	}
```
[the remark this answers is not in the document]
HBuczynski: App command can look as follow:
1. `app usb0 psh`
2. `app flash0 @psh`
3. `app flash0 psh(100:100)`

This loop is needed. In the first and second case presented above, only `cmdline` is passed in command, after that the `for` loop will be broken.
In the third case, all parameters will be parsed.

The `switch` statement can be optimized, I will push changes.

cmd.c
```diff
@@ -384,89 +375,81 @@ void cmd_write(char *s)
 }
 
 
-/* Function parse device name or parameters. There is distinction whether it is storage or external dev.
- * For external dev, name shall be provided. For storage device offset and size shall be provided. */
-#if 0
-static int cmd_parseDev(phfs_conf_t *dev, const char (*args)[LINESZ + 1], u16 *argsID, u16 cmdArgsc)
+static int cmd_parseDev(handler_t *h, addr_t *offs, size_t *sz, const char (*args)[LINESZ + 1], u16 *argsID, u16 argsc)
```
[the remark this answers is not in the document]
HBuczynski: The matrix has to be passed in this way as an argument to the function.

cmd.c
```diff
@@ -384,89 +375,81 @@ void cmd_write(char *s)
...
 {
-	if (plostd_ishex(args[++(*argsID)]) < 0) {
-		plostd_printf(ATTR_LOADER, "\nOpening device %s ..\n", cmd_common.devs[dev->dn].name);
-		dev->handle = phfs_open(dev->dn, args[(*argsID)], 0);
+	const char *alias = args[(*argsID)++];
+
+	if ((*argsID) >= argsc) {
```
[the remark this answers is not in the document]
HBuczynski: Done

cmd.c
HBuczynski: Yes, you have right. I have added information about failure.

cmd.c
HBuczynski: Done

cmd.c
HBuczynski: Done

hal/armv7a/zynq7000/cmd.c
HBuczynski: Done

agkaminski (approved):

PR #59, opened 2021-04-26
script: simplification of script processing

kemonats:
Instead of creating an array of copies of individual script lines, an array of pointers to those lines is created.
The file `script.plo.h` is created in the `Makefile` and then included in `script.c`.

Other changes:

- `script_expandAlias()`:  moved the calculation of name length and parameter position outside the for loop.
- ` Makefile`:  removed unnecessary semicolons.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,43 +20,51 @@ include hal/$(TARGET_SUFF)/$(TARGET_SUBFAMILY)/Makefile
 
 CFLAGS += $(BOARD_CONFIG)
 CFLAGS += -I../plo/hal/$(TARGET_SUFF)/$(TARGET_SUBFAMILY)
+CFLAGS += -I$(BUILD_DIR)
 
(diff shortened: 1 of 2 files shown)


agkaminski:
@pawelpisarczyk What are your thoughs about this approach?

There is an added benefit of not modifying the .elf file by ourselves (it is up to debate if this is a bad thing).
On the other hand the approach presented in this PR is also somewhat hacky (a nonorthodox usage of `#include`) and complicates a little the Makefile.

kemonats:
A similar use of `#include` is in [strerror.c](https://github.com/phoenix-rtos/libphoenix/blob/master/string/strerror.c)

HBuczynski:
Thank you for Your suggestion. However, in the new version of plo, there is a different approach for running scripts . Your changes can't be applied to current plo. 
PR is closed.

PR #60, opened 2021-04-28
plo-uart: use new devs interface 

HBuczynski:


devices/uart-imxrt117x/uart.c
```diff
@@ -93,15 +134,14 @@ int uart_getTXcount(uart_t *uart)
...
 {
 	uart_t *uart;
-	--pn;
 
-	if (pn >= (sizeof(uartConfig) / sizeof(uartConfig[0])) || !uartConfig[pn])
+	if (uart_checkInstance(dn--) <= 0)
 		return ERR_ARG;
 
-	uart = &uart_common.uarts[uartPos[pn]];
+	uart = &uart_common.uarts[dn];
```
agkaminski, line 144: As with 106x, I would avoid modifying dn for readabilty sake

PR #62, opened 2021-05-24
imxrt117x: Init all pads as slow slew-rate and low drive strength (EMC)

agkaminski:
JIRA: NIL-123

This should do for initializing all pins as slow and weak for EMC purpose. If someone need pin to be fast, they should reconfigure pin as they wish.

Not tested on HW yet, don't merge without verification. Might have bad effect on FlexSPI


agkaminski:
Tested on HW, no problem with FlexSPI has been observed

PR #71, opened 2021-05-28
cmds/app: simplification of the code

kemonats:
- corrected checking the number of arguments ( can be 4 or 5 )
- corrected length check of appname

### DIFF
--- a/cmds/app.c
+++ b/cmds/app.c
@@ -132,7 +132,7 @@ static int cmd_app(char *s)
 		syspage_showApps();
 		return EOK;
 	}
-	else if (argsc < 4 || argsc > 6) {
+	else if (argsc < 4 || argsc > 5) {
 		log_error("\nWrong args: %s", s);
 		return -EINVAL;
 	}
@@ -153,8 +153,8 @@ static int cmd_app(char *s)
 	}
 
 	/* ARG_2: cmdline */
-	if (argID >= argsc) {
-		log_error("\nWrong args: %s", s);
+	if (argID != (argsc - 3)) {
+		log_error("\nInvalid args, 'dmap' is not declared");
 		return -EINVAL;
 	}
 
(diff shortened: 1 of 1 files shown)


cmds/app.c
```diff
@@ -153,7 +153,7 @@ static int cmd_app(char *s)
 	}
 
 	/* ARG_2: cmdline */
-	if (argID >= argsc) {
+	if (argID != (argsc - 3)) {
 		log_error("\nWrong args: %s", s);
 		return -EINVAL;
 	}
```
HBuczynski, line 159: I think that the whole condition can be removed.
If there will be wrong number of arguments it will be handled at the beginning of the function.

HBuczynski (changes requested):

cmds/app.c
kemonats: This condition is needed for such a case:
```
app dev -x name imap
```
`argsc == 4` but `dmap` is missing.

HBuczynski: You have right, it will not be checked at the beginning of the function. However, this condition will be handled by the following function and information that `dmap` or `imap` don't exist, will be printed:
https://github.com/phoenix-rtos/plo/blob/3db025075842cf5144538f412c670bd75521481c/cmds/app.c#L106-L109

https://github.com/phoenix-rtos/plo/blob/3db025075842cf5144538f412c670bd75521481c/syspage.c#L612-L615

So this check is redundant. Your case will be handled and appropriate information will be printed.

kemonats: I meant `dmap` with help and not `dmap` as a variable.
```
command line:
app dev -x name imap
argsc == 4
argID value
0     dev
1     -x
2     name
3     imap
4     ???? - only 4 arguments
```
and after lines:
```
hal_memcpy(dmap, args[++argID], sizeof(dmap));
dmap[sizeof(dmap) - 1] = '\0';
```
What is the value of the `dmap` variable.
And it will print a strange message.

HBuczynski: The reason why I wanted to print error's message from sysapge is that it is more precise. User knows that map declaration is wrong. I think that we can find a trade-off if you write better log_error. Smth like this:
```suggestion
	if (argID != (argsc - 3)) {
		log_error("\nInvalid args, 'dmap' is not declared");
		return -EINVAL;
	}
```

kemonats:
Done

PR #73, opened 2021-05-28
cmds/echo: simplification of the code

kemonats:
- added argument conversion to lowercase
- changed `hal_strcmp()` to simple check

### DIFF
--- a/cmds/echo.c
+++ b/cmds/echo.c
@@ -16,6 +16,7 @@
 #include "cmd.h"
 
 #include <lib/log.h>
+#include <lib/ctype.h>
 #include <hal/hal.h>
 
 
@@ -29,6 +30,7 @@ static int cmd_echo(char *s)
 {
 	unsigned int argsc;
 	cmdarg_t *args;
+	char *p;
 
 	argsc = cmd_getArgs(s, DEFAULT_BLANKS, &args);
 	/* Show echo status */
@@ -48,11 +50,18 @@ static int cmd_echo(char *s)
 		return -EINVAL;
 	}
 
+	/* convert to lowercase */
+	p = args[0];
+	for (p = args[0]; *p; p++) {
+		if (isupper(*p))
+			*p |= 0x20;
+	}
+
 	/* Set echo */
-	if (hal_strcmp(args[0], "ON") == 0 || hal_strcmp(args[0], "on") == 0) {
+	if ((args[0][0] == 'o') && (args[0][1] == 'n') && (args[0][2] == '\0')) {
 		log_setEcho(1);
 	}
-	else if (hal_strcmp(args[0], "OFF") == 0 || hal_strcmp(args[0], "off") == 0) {
+	else if ((args[0][0] == 'o') && (args[0][1] == 'f') && (args[0][2] == 'f') && (args[0][3] == '\0')) {
 		log_setEcho(0);
 	}
 	else {


HBuczynski:
I don't think so that You simplified the code.

1. You provide additional variable.
2. You add extra code `/* convert to lowercase */`. The only change which should be added is that `args` should be const variable  and in the rest of the code ;)
3. You removed `lib_strcmp`. Your proposition don't simplified the code but make it less readable.

kemonats:
My suggestion is because the `lib_strcmp()` function is type `inline` and simplification is in terms of compiler generated code.
Personally, I see no argument for this function and the other `lib_str*()` to be type `inline` and written in inline assembly.
In the current version with the `-O2` optimization flag, the code of the `lib_strcmp()` function is inserted 4 times.
This results in an increase code size.
Even with `-Os` flag ( only one copy of the `lib_strcmp()` code is inserted ) my proposal creates a shorter code.

And so the code size of `echo.o`:
currebt version
```
zynq7000 (-Os)   386
imxrt106x (-O2)  505
imxrt117x (-O2)  505
```
my proposal
```
zynq7000 (-Os)   330
imxrt106x (-O2)  346
imxrt117x (-O2)  346
```

Converting to lowercase allow only on/off to be checked.
If both on/off and ON/OFF are allowed, then you need to change the help.
Another solution is to use only on/off, I suggest this option.

HBuczynski:
Well, if the problem is with inline functions, this suggestion should be notified in `hal.` In this case there is a proper usage of `hal_strcmp` (the hal causes larger code's size) and code is more readable. 

Ok, I can agree to use only lowercase.

HBuczynski:
Using only lowercase was added in the [PR](https://github.com/phoenix-rtos/plo/pull/77).

cmds/echo.c
```diff
@@ -48,11 +50,18 @@ static int cmd_echo(char *s)
...
+	/* convert to lowercase */
+	p = args[0];
+	for (p = args[0]; *p; p++) {
+		if (isupper(*p))
+			*p |= 0x20;
+	}
+
 	/* Set echo */
-	if (hal_strcmp(args[0], "ON") == 0 || hal_strcmp(args[0], "on") == 0) {
+	if ((args[0][0] == 'o') && (args[0][1] == 'n') && (args[0][2] == '\0')) {
```
gerard5, line 61: I'm not delving into the `cmd_getArgs()` implementation, but is that ok that the length of the string `args[0]` is not checked here as in the `hal_strcmp()` case it is ? Here only the presence of `\0` at third (assumed) byte is checked, what if the size of `args[0]` < 3 ?

cmds/echo.c
```diff
@@ -48,11 +50,18 @@ static int cmd_echo(char *s)
...
 		log_setEcho(1);
 	}
-	else if (hal_strcmp(args[0], "OFF") == 0 || hal_strcmp(args[0], "off") == 0) {
+	else if ((args[0][0] == 'o') && (args[0][1] == 'f') && (args[0][2] == 'f') && (args[0][3] == '\0')) {
```
gerard5, line 64: Similarly as above.

cmds/echo.c
```diff
@@ -48,11 +50,18 @@ static int cmd_echo(char *s)
 		return -EINVAL;
 	}
 
+	/* convert to lowercase */
+	p = args[0];
```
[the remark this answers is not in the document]
kemonats: You are right. I changed from a `while` loop to a `for` loop and didn't remove that line.

cmds/echo.c
kemonats: e.g.
```
command line: echo o
args[0][0] == 'o'  true
args[0][1] == 'n'  false ( '\0' character )
```
and the whole condition is `false`

kemonats:
Of course using `strcmp()` is more readable.
In this PR I wanted to show how using `inline` functions affects code size, especially in the case of -O2 optimization.

cmds/echo.c
gerard5: Exactly, that's the expected answer.

cmds/echo.c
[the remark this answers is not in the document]
gerard5: I thought so.

HBuczynski:
> Of course using `strcmp()` is more readable.
> In this PR I wanted to show how using `inline` functions affects code size, especially in the case of -O2 optimization.
> To close this PR

Thanks for paying attention for optimization aspects, you can create an issue related to hal's functions to discuss it.  
However your PR's title is **simplification of the code** as you have noticed yourself, your suggestion focused on optimization rather than simplification of the code.
For this reason, I would like to ask you to close this PR.

PR #79, opened 2021-06-01
plo: fix uart

HBuczynski:
This PR should solve the problem with UART driver which causes that plo occasionally freezes on imxrt106x.
Within this changes the volatile types are added to variables which values are changed in interrupts.

@jsarzy, please test this solution with runner.


jsarzy:
I tested it and it seems to work, but I didn't verify correctness of the code.

PR #80, opened 2021-06-07
flash-imxrt: fix writing to flash after initialization by bootrom

HBuczynski:
Only part of the `_fcfb` were copied to the flash's configuration structure, so that `page_size` and `sector_size` were 0. For this reason, `write` and `erase` operations haven't worked.

Moreover, `_fcfb` for imxrt106x has been fixed.

**NOTE:** This PR has an impact on imxrt117x. This change has not been tested on this platform.


agkaminski (approved):
LGTM

PR #83, opened 2021-06-08
cmds/cmd: command line parsing changed

kemonats:
The function `cmd_parse()` is called in functions:
```
cmd_call()
cmd_prompt()
cmd_run()
```
The `'\n'` character occurs in the argument passed to `cmd_parse()` only in `cmd_run()`.
In this PR parsing `'\n'` character has been moved to `cmd_run()`.
The `cmd_parse()` function type has also been changed.

### DIFF
--- a/cmds/call.c
+++ b/cmds/call.c
@@ -78,7 +78,10 @@ static int cmd_call(char *s)
 		offs += res;
 		if (res == 0 || c == '\n' || i == (sizeof(buff) - 1)) {
 			buff[i] = '\0';
-			cmd_parse(buff);
+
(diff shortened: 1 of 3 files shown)


HBuczynski:
In the last PR: https://github.com/phoenix-rtos/plo/pull/97 `cmd_parse()` has been changed so that your changes are no longer valid. Sorry that your PR has not been merged so far but  we had to add some changes in commands.

Moreover, the possibility of transferring the entire script in `cmd_run()` to `cmd_parse()` is a good feature which should not be changed.

PR #84, opened 2021-06-08
cmd/kernel: add optional kernel alias/name argument

gerard5:
Two improvements within this pull request:
- add the ability to indicate an alias name to the `kernel` command like e.g. `kernel flash1 my-kernel` instead of hard-coded name `phoenix-armv7m7-imxrt106x.elf`.
- add kernel address validation in the `go` command if the kernel was properly loaded
- and a few minor fixes

JIRA: RTOS-43, RTOS-44, PD-89, PD-117

### DIFF
--- a/cmds/app.c
+++ b/cmds/app.c
@@ -27,7 +27,7 @@ static void cmd_appInfo(void)
 }
 
 
-static int cmd_cpphfs2map(handler_t handler, const char *imap)
+static int cmd_cpphfs2map(handler_t handler, const char *imap, void *top)
 {
 	int res;
 	addr_t offs = 0;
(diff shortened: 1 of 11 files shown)


cmds/app.c
```diff
@@ -30,17 +30,23 @@ static void cmd_appInfo(void)
 static int cmd_cpphfs2map(handler_t handler, const char *imap)
 {
 	int res;
-	addr_t offs = 0;
+	addr_t top, offs = 0;
 	u8 buff[SIZE_MSG_BUFF];
 
+	if ((res = syspage_getMapTop(imap, &top)) < 0)
```
HBuczynski, line 36: `syspage_getMapTop` is invoked here:
https://github.com/phoenix-rtos/plo/blob/baefaa9deb107d483d23bf48db1164819baae9f0/cmds/app.c#L82-L85
I propose to pass top value as argument to `cmd_cpphfs2map`.

HBuczynski (changes requested):

cmds/app.c
gerard5: Done.

PR #96, opened 2021-06-17
cmd/copy: register alias to file

HBuczynski:
## Description

Copy command registers alias to file.

## Motivation and Context

When user wants to copy program or file to storage, the alias to file will be added automatically.
JIRA: PD-124

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt106x).

## Checklist:

## Special treatment

### DIFF
--- a/cmds/copy.c
+++ b/cmds/copy.c
@@ -29,42 +29,40 @@ static void cmd_copyInfo(void)
 
 static ssize_t cmd_cpphfs2phfs(handler_t srcHandler, addr_t srcAddr, size_t srcSz, handler_t dstHandler, addr_t dstAddr, size_t dstSz)
 {
-	int res;
-	addr_t addr = dstAddr;
-	size_t size, chunk;
+	ssize_t res;
 	u8 buff[SIZE_MSG_BUFF];
+	size_t chunk, rsz = 0, wsz = 0;
 
 	/* Size is not defined, copy the whole file                 */
 	if (srcSz == 0 && dstSz == 0)
-		size = -1;
+		srcSz = -1;
 	/* Size is defined, use smaller one to copy piece of memory */
 	else if (srcSz != 0 && dstSz != 0)
-		size = (srcSz < dstSz) ? srcSz : dstSz;
+		srcSz = (srcSz < dstSz) ? srcSz : dstSz;
 	/* One of the size is not defined, use the defined one      */
 	else
-		size = (srcSz > dstSz) ? srcSz : dstSz;
+		srcSz = (srcSz > dstSz) ? srcSz : dstSz;
 
 	do {
-		if (size > sizeof(buff))
-			chunk = sizeof(buff);
-		else
-			chunk = size;
-
-		if ((res = phfs_read(srcHandler, srcAddr, buff, chunk)) < 0) {
+		chunk = ((srcSz - rsz) > sizeof(buff)) ? sizeof(buff) : (srcSz - rsz);
+		if ((res = phfs_read(srcHandler, srcAddr + rsz, buff, chunk)) < 0) {
 			log_error("\nCan't read data\n");
-			return -EIO;
+			return res;
 		}
-		srcAddr += res;
-		size -= res;
+		rsz += res;
 
-		if ((res = phfs_write(dstHandler, addr, buff, res)) < 0) {
-			log_error("\nCan't write data to address: 0x%x\n", addr);
-			return -EIO;
+		chunk = res;
+		while (chunk) {
+			if ((res = phfs_write(dstHandler, dstAddr + wsz, buff, chunk)) < 0) {
+				log_error("\nCan't write data to address: 0x%x\n", dstAddr + wsz);
+				return res;
+			}
+			chunk -= res;
+			wsz += res;
 		}
-		addr += res;
-	} while (size > 0 && res > 0);
+	} while ((srcSz - rsz) > 0 && res > 0);
 
-	return addr - dstAddr;
+	return wsz;
 }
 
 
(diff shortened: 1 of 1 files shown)


cmds/copy.c
```diff
@@ -27,9 +27,10 @@ static void cmd_copyInfo(void)
 }
 
 
-static int cmd_cpphfs2phfs(handler_t srcHandler, addr_t srcAddr, size_t srcSz, handler_t dstHandler, addr_t dstAddr, size_t dstSz)
+static ssize_t cmd_cpphfs2phfs(handler_t srcHandler, addr_t srcAddr, size_t srcSz, handler_t dstHandler, addr_t dstAddr, size_t dstSz)
 {
 	int res;
+	addr_t addr = dstAddr;
```
agkaminski, line 33: You have 3 variables that you modify on each pass of the loop. Maybe using one variable (which would represent current position in buffers), adding this value to addresses and comparing it with const size would be better? You wouldn't need to calculate retval this way and IMHO code would be easier to read and less prone to errors.

HBuczynski: Yeah, it can be simplified. However, the `res` value might differ from read and write value, it is the worst scenario. 
For this reason I decided to use more than one variable.

agkaminski: Yeah, good point. What if res is different for read and write? E.g. we read 20 bytes and manage to write only 10. Wouldn't we silently drop 10 bytes of data? I think that write should have it's own loop to write all read data before next read operation.

agkaminski: Or we can read again from offset we got from write - either way, one indexing variable is better ;)

agkaminski (changes requested):
`cmd_cpphfs2phfs` needs rework before merge

cmds/copy.c
```diff
@@ -127,15 +132,15 @@ static int cmd_copy(char *s)
...
 		return -EINVAL;
 
-	if (cmd_parseDev(&h[1], &offs[1], &sz[1], args, &argsID, argsc) < 0)
+	if (cmd_parseDev(&h[1], &offs[1], &sz[1], args, &argsID, argsc, &file[1]) < 0)
 		return -EINVAL;
 
 	/* Copy data between devices */
 	log_info("\nCopying data, please wait...");
-	if (cmd_cpphfs2phfs(h[0], offs[0], sz[0], h[1], offs[1], sz[1]) < 0) {
+	if ((res = cmd_cpphfs2phfs(h[0], offs[0], sz[0], h[1], offs[1], sz[1])) < 0) {
```
gerard5, line 143: You may `return res` instead of` return -EINVAL` below when the `res` already contains a negative result, such as` EIO`, which is more descriptive than `EINVAL` in this case. This note also applies to the lines above which contain `if (cmd_parseDev (...) < 0)`, You can use `res` and return the cause of the error. What do you think about it ?

cmds/copy.c
```diff
@@ -56,18 +57,18 @@ static int cmd_cpphfs2phfs(handler_t srcHandler, addr_t srcAddr, size_t srcSz, h
 		srcAddr += res;
 		size -= res;
 
-		if ((res = phfs_write(dstHandler, dstAddr, buff, res)) < 0) {
-			log_error("\nCan't write data to address: 0x%x\n", dstAddr);
+		if ((res = phfs_write(dstHandler, addr, buff, res)) < 0) {
```
gerard5, line 60: You may `return res` instead of `return -EIO` below, I think `phfs_write()` should tell the real cause of failure. Especially in the context that `cmd_parse(..)` will return `int` error cause in a few days.

cmds/copy.c
```diff
@@ -115,8 +118,10 @@ static int cmd_parseDev(handler_t *h, addr_t *offs, size_t *sz, cmdarg_t *args,
 static int cmd_copy(char *s)
 {
 	size_t sz[2];
+	ssize_t res;
 	addr_t offs[2];
 	handler_t h[2];
+	const char *file[2];
```
kemonats, line 124: Wouldn't it be better to define a structure
e.g.
```
typedef struct{
	size_t sz;
	addr_t offs;
	handler_t h;
	const char *file;
} copy_dev_t;
```
In `cmd_copy()` define the variables
```
copy_dev_t src;
copy_dev_t dst;
```
and then 
```
static ssize_t cmd_cpphfs2phfs(copy_dev_t *src, copy_dev_t *dst)
static int cmd_parseDev(copy_dev_t *dev, cmdarg_t *args, unsigned int *argsID, unsigned int argsc)
```
This will reduce the number of arguments in functions:
`cmd_cpphfs2phfs()` from 6 to 2
`cmd_parseDev()` from 7 to 4

What do you think about it ?

cmds/copy.c
kemonats, line 143: If you are using `res` variable here then you can simplify the code to:
```
	res = cmd_cpphfs2phfs(h[0], offs[0], sz[0], h[1], offs[1], sz[1]);

	phfs_close(h[0]);
	phfs_close(h[1]);

	if (res < 0) {
		log_error("\nCopying failed");
		return -EINVAL;
	}

	log_info("\nFinished copying");
```

cmds/copy.c
agkaminski: @pawelpisarczyk I know that we try to minimalize number of structures and typedefs to the minimum, but it might be justified in above case. In the above code are arrays of 2 elements of members of such proposed structure. Defining it (as struct copyDev or typedef copyDev_t) IMHO would code readability.

HBuczynski: Yeah, the reason why I decided to create arrays of 2 elements to reduce the instances of additional structures and typedefs as it is one of the our rules. 
I agree that creating structure in this case will increase to code readability.

cmds/copy.c
HBuczynski: Done

cmds/copy.c
HBuczynski: Done

cmds/copy.c
HBuczynski: done

cmds/copy.c
```diff
@@ -29,42 +29,40 @@ static void cmd_copyInfo(void)
...
+		srcSz = (srcSz > dstSz) ? srcSz : dstSz;
 
 	do {
-		if (size > sizeof(buff))
-			chunk = sizeof(buff);
-		else
-			chunk = size;
-
-		if ((res = phfs_read(srcHandler, srcAddr, buff, chunk)) < 0) {
+		chunk = ((srcSz - rsz) > sizeof(buff)) ? sizeof(buff) : srcSz;
```
agkaminski, line 47: Should be:
`chunk = ((srcSz - rsz) > sizeof(buff)) ? sizeof(buff) : (srcSz - rsz);`
Right now the last read will have size of all others + `(srcSz - rsz)`;

HBuczynski: Yeah, you have right. I overlooked it. Fixed

agkaminski (approved):
I think that the idea o copyDev_t structure can be implemented in the future PR, if it is decided to. For now this PR LGTM.

PR #97, opened 2021-06-18
Change the script and command parser to accept arguments like in C  `main(argc, argv)`

gerard5:
## Description

This PR introduces a change to the way arguments are passed to plo command line programs, the convention introduced in this PR is similar to the standard C `main` function, arguments are passed to commands using `int argc, const char *argv[]` construct. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7m7-imxrt106x).

## Checklist:

## Special treatment

### DIFF
--- a/cmds/app.c
+++ b/cmds/app.c
@@ -29,24 +29,24 @@ static void cmd_appInfo(void)
 
 static int cmd_cpphfs2map(handler_t handler, const char *imap, addr_t top)
 {
-	int res;
+	ssize_t len, res;
 	addr_t offs = 0;
 	u8 buff[SIZE_MSG_BUFF];
 
 	do {
-		if ((res = phfs_read(handler, offs, buff, SIZE_MSG_BUFF)) < 0) {
+		if ((len = phfs_read(handler, offs, buff, SIZE_MSG_BUFF)) < 0) {
 			syspage_setMapTop(imap, top);
 			log_error("\nCan't read data");
-			return -EIO;
+			return len;
 		}
 
-		if (syspage_write2Map(imap, buff, res) < 0) {
+		offs += len;
+
+		if ((res = syspage_write2Map(imap, buff, len)) < 0) {
 			syspage_setMapTop(imap, top);
-			return -EINVAL;
+			return res;
 		}
-
-		offs += res;
-	} while (res > 0);
+	} while (len > 0);
 
 	return EOK;
 }
@@ -54,7 +54,7 @@ static int cmd_cpphfs2map(handler_t handler, const char *imap, addr_t top)
 
 static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const char *dmap, const char *cmdline, u32 flags)
 {
-	int res;
+	size_t res;
 	Elf32_Ehdr hdr;
 	addr_t start, end;
 	addr_t addr, offs = 0;
@@ -63,7 +63,7 @@ static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const c
 	/* Check ELF header */
 	if ((res = phfs_read(handler, offs, (u8 *)&hdr, (u32)sizeof(Elf32_Ehdr))) < 0) {
 		log_error("\nCan't read data");
-		return -EIO;
+		return res;
 	}
 
 	if ((hdr.e_ident[0] != 0x7f) || (hdr.e_ident[1] != 'E') || (hdr.e_ident[2] != 'L') || (hdr.e_ident[3] != 'F')) {
@@ -83,7 +83,7 @@ static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const c
 
 	if ((res = phfs_map(handler, offs, size, mAttrRead | mAttrWrite, start, size, attr, &addr)) < 0) {
 		log_error("\nDevice is not mappable in %s", imap);
-		return -EINVAL;
+		return res;
 	}
 
 	/* Copy program's elf to imap */
@@ -95,8 +95,8 @@ static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const c
 		syspage_getMapTop(imap, &end);
 	}
 	else if (res == dev_isMappable) {
-		if (phfs_getFileAddr(handler, &offs) < 0)
-			return -EINVAL;
+		if ((res = phfs_getFileAddr(handler, &offs)) < 0)
+			return res;
 
 		start = offs + addr;
 		end = start + size;
@@ -106,9 +106,9 @@ static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const c
 		return -EIO;
 	}
 
-	if (syspage_addProg(start, end, imap, dmap, cmdline, flags) < 0) {
+	if ((res = syspage_addProg(start, end, imap, dmap, cmdline, flags)) < 0) {
 		log_error("\nCan't add program to syspage");
-		return -EINVAL;
+		return res;
 	}
 
 	return EOK;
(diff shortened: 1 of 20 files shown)


gerard5:
Previously [df023e3](https://github.com/phoenix-rtos/plo/pull/97/commits/df023e3a1dcd642b5ed054f408cf2b2c7f8eac6b) pushed version assumed this construct `int argc, const char *argv[]` as an argument list (which worked fine) and is more C99, but I changed my mind so that each command would take the argument list either as either `int argc, char *argv[]` or `int argc, char **argv`. This is because it complies with C89 and how most of our software is defined, so I "push-force" changes like that, if it requires discussion I am open to.

cmds/cmd.c
```diff
@@ -28,61 +29,53 @@
...
-	while ((c = line[*pos]) != 0) {
+		while (isgraph(**lines) && buf < end)
+			*buf++ = *(*lines)++;
 
-		/* Test separators */
-		for (i = 0; blanks[i]; i++) {
-			if (c == blanks[i])
-				break;
+		if (buf >= end) {
+			log_error("\ncmd: Command buffer to small");
```
agkaminski, line 69: too small

cmds/console.c
```diff
@@ -26,32 +26,31 @@ static void cmd_consoleInfo(void)
...
 		return -EINVAL;
 	}
 
+
 	/* Get major/minor */
-	major = lib_strtoul(args[0], &endptr, 0);
-	if (*endptr) {
-		log_error("\nWrong major value: %s", args[0]);
+	major = lib_strtoul(argv[1], &endptr, 0);
+	if (*endptr != '.') {
```
agkaminski, line 46: Why did it work before?

agkaminski (commented):
I don't know about the general idea - do we really need command name in `argv[0]` when we are already in a function to parse this command (so we know what command is being executed anyway).

Implementation (apart from small typo) LGTM

cmds/cmd.c
```diff
@@ -28,61 +29,53 @@
...
+			else
+				break;
+		}
 
-/* TODO: old code needs to be cleaned up */
-static char *cmd_getnext(const char *line, unsigned int *pos, const char *blanks, char *word, unsigned int len)
-{
-	char c;
-	unsigned int i, wp = 0;
+		if (argc + 1 >= argvsz) {
```
HBuczynski, line 58: This condition allows to have one argument less than the maximum size is defined. For example, `map` should consists of 5 arguments: `map itcm 0x0 0x40000 rwx`. When I have set `MAX_CMD_ARGS_NB = 5`, the parsing has stopped on `map` command.

HBuczynski (changes requested):
I have tested your changes on hardware. I tried to use `call` command but it doesn't work after your changes. Have you checked this command as well ?

cmds/console.c
gerard5: Previously, there were `cmd_getArgs(s, ". \t", &args)` that tokenized the string of arguments, but the ` ` and `\t` separators have already been removed in `cmd_parse()`, so ` ` `\t` characters never occurred, but the only `.` character, so `cmd_getArgs` split the string with using `.` delimiter into two strings. It worked, but I think the proposed solution in this PR  is clearer.

cmds/app.c
```diff
@@ -106,63 +106,61 @@ static int cmd_loadApp(handler_t handler, size_t size, const char *imap, const c
...
 
 
 /* TODO: allow to add more than 2 maps */
-static int cmd_app(char *s)
+static int cmd_app(int argc, char *argv[])
 {
-	int argID = 0;
-	unsigned int argsc, pos;
-	cmdarg_t *args;
+	int res, argID = 0;
```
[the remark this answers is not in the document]
gerard5: Done.

cmds/copy.c
```diff
@@ -66,74 +66,74 @@ static ssize_t cmd_cpphfs2phfs(handler_t srcHandler, addr_t srcAddr, size_t srcS
...
 
 	/* Open device using alias to file */
 	if (*endptr) {
 		*offs = 0;
 		*sz = 0;
-		*file = args[*argsID];
-		if (phfs_open(alias, args[*argsID], 0, h) < 0) {
-			log_error("\nCan't open file '%s' on %s", args[*argsID], alias);
+		*file = argv[*argvID];
+		if (phfs_open(alias, argv[*argvID], 0, h) < 0) {
```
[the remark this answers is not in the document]
gerard5: Done.

cmds/cmd.c
```diff
@@ -28,61 +29,53 @@
...
 
-	while (line[*pos]) {
-		for (p = blanks; *p && line[*pos] != *p; p++);
-		if (!*p)
-			break;
-		(*pos)++;
-	}
-}
+	if (lines == NULL || *lines == NULL || **lines == '\0')
+		return -1; /* EOF */
```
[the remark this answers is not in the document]
gerard5: This is special case, as denoted in comment EOF isn't an error, but `end of file`, as we've talked about.

cmds/cmd.c
gerard5: Yes, `MAX_CMD_ARGS_NB` shall be  increased by `+1`, this is due to the argument lis termination with `NULL`, this is from standard C, where You could count arguments till `NULL`. So, I'll increase by  +1 `MAX_CMD_ARGS_NB` marking that last argument `+1` is `NULL`, as we talked.
```
#define MAX_CMD_ARGS_NB (10 + 1)
```

cmds/cmd.c
gerard5: Corrected, thank You.

cmds/cmd.c
HBuczynski: Ok. Can you write a comment in the code why `+1` is added to avoid doubts in the future?

kemonats: Isn't it simpler to do it this way:

cmd.h
```
#define MAX_CMD_ARGS_NB 10
```
cmd.c
```
int cmd_parse(const char *script)
{
	char *argv[MAX_CMD_ARGS_NB + 1]; // and even +2 (cmd nane + NULL)

		argc = cmd_parseArgLine(&script, argline, SIZE_CMD_ARG_LINE, argv, sizeof(argv) / sizeof(argv[0]);

```

gerard5: @HBuczynski , changed name of MAX_CMD_ARGS_NB to SIZE_CMD_ARGV to denote array size, also added comments where apropriate.
@kemonats +1, +2 .. is not the case. We should be sure of the length of the array everywhere, so `cmd_parseArgLine()` checks `(argc + 1) > SIZE_CMD_ARGV` , thus `argv` should be at least one for terminating `NULL` pointer, then for this edge case there will be no space for arguments only terminating `NULL` pointer will be there - thats valid case. And. If `SIZE_CMD_ARGV` is zero then `cmd_parseArgLine()` should return negative `EINVAL`.

HBuczynski (approved):

HBuczynski:
LGTM

cmds/cmd.c
kemonats: In cmd_parseArgLine() checks (argc + 1) > argvsz
```
static int cmd_parseArgLine(const char **lines, char *buf, size_t bufsz, char *argv[], size_t argvsz)

argc = cmd_parseArgLine(&script, argline, SIZE_CMD_ARG_LINE, argv, sizeof(argv) / sizeof(argv[0]);
```
and so argvsz equals the number of elements of the argv array.

gerard5: Exactly. `SIZE_CMD_ARGV` tells how many pointers are in argv[], but the last pointer is reserved and always points to  `NULL` , thus +1 to indicate that fact, see comments there. It `(argc + 1) > argvsz` is not actual, see merged version:
https://github.com/phoenix-rtos/plo/blob/d9fefe2330d9aa771825db235a1826565534b864/cmds/cmd.c#L58-L60

PR #106, opened 2021-07-01
Clean up standard functions interface

Maxez:
## Description

Changed plo standard functions signatures to resemble their standard C library equivalents.

## Motivation and Context

The interface is cleaner. Some type casting is no longer necessary.
[JIRA: PD-141](https://jira.phoenix-rtos.com/browse/PD-141)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a9-zynq7000, armv7m7-imxrt106x (checked for build errors and warnings).

## Checklist:

## Special treatment

### DIFF
--- a/devices/flash-imxrt/flashdrv.c
+++ b/devices/flash-imxrt/flashdrv.c
@@ -179,7 +179,7 @@ static int flashdrv_defineFlexSPI(flash_context_t *ctx)
 
 
 /* Device interface */
-static ssize_t flashdrv_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int len, unsigned int timeout)
+static ssize_t flashdrv_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
 {
 	if (minor >= FLASH_NO)
 		return -EINVAL;
@@ -189,7 +189,7 @@ static ssize_t flashdrv_read(unsigned int minor, addr_t offs, u8 *buff, unsigned
 }
 
 
-static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
+static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const void *buff, size_t len)
 {
 	addr_t bOffs = 0;
 	flash_context_t *ctx;
@@ -206,7 +206,7 @@ static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const u8 *buff, u
 
 	while (size) {
 		if (size < buffSz) {
-			hal_memcpy(flashdrv_common.buff, buff + bOffs, size);
+			hal_memcpy(flashdrv_common.buff, (const char *)buff + bOffs, size);
 			hal_memset(flashdrv_common.buff + size, 0xff, buffSz - size);
 			if (flashdrv_bufferedPagesWrite(ctx, offs + bOffs, (const char *)flashdrv_common.buff, buffSz) < 0)
 				return -EINVAL;
@@ -214,7 +214,7 @@ static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const u8 *buff, u
 			break;
 		}
 
-		if (flashdrv_bufferedPagesWrite(ctx, offs + bOffs, (const char *)(buff + bOffs), buffSz) < 0)
+		if (flashdrv_bufferedPagesWrite(ctx, offs + bOffs, (const char *)buff + bOffs, buffSz) < 0)
 			return -EINVAL;
 
 		bOffs += buffSz;
--- a/devices/uart-imxrt106x/uart.c
+++ b/devices/uart-imxrt106x/uart.c
@@ -434,9 +434,9 @@ static void uart_initPins(void)
 
 /* Device interface */
 
-static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int len, unsigned int timeout)
+static ssize_t uart_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
 {
-	u16 l, cnt;
+	size_t l, cnt;
 	uart_t *uart;
 	time_t start;
 
@@ -459,7 +459,7 @@ static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int
 	hal_memcpy(buff, &uart->rxBuff[uart->rxTail], l);
 	cnt = l;
 	if ((len > l) && (uart->rxHead < uart->rxTail)) {
-		hal_memcpy(buff + l, &uart->rxBuff[0], min(len - l, uart->rxHead));
+		hal_memcpy((char *)buff + l, &uart->rxBuff[0], min(len - l, uart->rxHead));
 		cnt += min(len - l, uart->rxHead);
 	}
 	uart->rxTail = ((uart->rxTail + cnt) % BUFFER_SIZE);
@@ -470,9 +470,9 @@ static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int
 }
 
 
-static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
+static ssize_t uart_write(unsigned int minor, const void *buff, size_t len)
 {
-	unsigned int l, cnt = 0;
+	size_t l, cnt = 0;
 	uart_t *uart;
 
 	if ((uart = uart_getInstance(minor)) == NULL)
@@ -490,7 +490,7 @@ static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
 	hal_memcpy(&uart->txBuff[uart->txTail], buff, l);
 	cnt = l;
 	if ((len > l) && (uart->txTail >= uart->txHead)) {
-		hal_memcpy(uart->txBuff, buff + l, min(len - l, uart->txHead));
+		hal_memcpy(uart->txBuff, (const char *)buff + l, min(len - l, uart->txHead));
 		cnt += min(len - l, uart->txHead);
 	}
 
@@ -511,15 +511,14 @@ static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
 }
 
 
-static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
+static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const void *buff, size_t len)
 {
-	unsigned int l;
+	ssize_t res;
+	size_t l;
 
-	for (l = 0; len;) {
-		if ((l = uart_write(minor, buff, len)) < 0)
+	for (l = 0; l < len; l += res) {
+		if ((res = uart_write(minor, (const char *)buff + l, len - l)) < 0)
 			return -ENXIO;
-		buff += l;
-		len -= l;
 	}
 
 	return len;
--- a/devices/uart-imxrt117x/uart.c
+++ b/devices/uart-imxrt117x/uart.c
@@ -450,10 +450,10 @@ static void uart_initPins(void)
 
 /* Device interface */
 
-static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int len, unsigned int timeout)
+static ssize_t uart_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
 {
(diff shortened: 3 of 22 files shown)


syspage.c
```diff
@@ -346,25 +346,25 @@ int syspage_save(void)
...
 	syspage_common.syspage->progssz = syspage_common.progsCnt;
 	syspage_common.syspage->syspagesz += (syspage_common.progsCnt * sizeof(syspage_program_t));
-	hal_memcpy((void *)(syspage_common.syspage->progs), syspage_common.progs, syspage_common.progsCnt * sizeof(syspage_program_t));
+	hal_memcpy(syspage_common.syspage->progs, syspage_common.progs, syspage_common.progsCnt * sizeof(syspage_program_t));
 
 	/* Save maps */
 	syspage_common.syspage->maps = (syspage_map_t *)((void *)syspage_common.syspage + syspage_common.syspage->syspagesz);
 	for (i = 0; i < syspage_common.mapsCnt; ++i)
-		hal_memcpy((void *)((void *)syspage_common.syspage->maps + i * sizeof(syspage_map_t)), &syspage_common.maps[i].map, sizeof(syspage_map_t));
+		hal_memcpy(syspage_common.syspage->maps + i * sizeof(syspage_map_t), &syspage_common.maps[i].map, sizeof(syspage_map_t));
```
agkaminski, line 360: That's not good, when you add to pointer of syspage_map_t  * type, you add i * sizeof(syspage_map_t ). So you do here:
syspage_common.syspage->maps + (i * sizeof(syspage_map_t) * sizeof(syspage_map_t).

Use (char *) for pointer arithmetic or just do:
syspage_common.syspage->maps + i or &syspage_common.syspage->maps[i]

agkaminski (changes requested):

syspage.c
Maxez: Thanks for spotting this bug - it's fixed now.

devices/flash-imxrt/flashdrv.c
```diff
@@ -189,7 +189,7 @@ static ssize_t flashdrv_read(unsigned int minor, addr_t offs, u8 *buff, unsigned
 }
 
 
-static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
+static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const void *buff, size_t len)
```
agkaminski, line 192: Now there are two instances of void pointer arithmetic in this function

devices/uart-imxrt106x/uart.c
```diff
@@ -434,9 +434,9 @@ static void uart_initPins(void)
 
 /* Device interface */
 
-static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int len, unsigned int timeout)
+static ssize_t uart_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
```
agkaminski, line 437: Now there are now instances of void pointer arithmetic in this function

devices/uart-imxrt106x/uart.c
```diff
@@ -470,9 +470,9 @@ static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int
 }
 
 
-static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
+static ssize_t uart_write(unsigned int minor, const void *buff, size_t len)
```
agkaminski, line 473: Now there are now an instance of void pointer arithmetic in this function

devices/uart-imxrt106x/uart.c
```diff
@@ -511,9 +511,9 @@ static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
 }
 
 
-static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
+static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const void *buff, size_t len)
```
agkaminski, line 514: Now there are now an instance of void pointer arithmetic in this function

devices/uart-imxrt117x/uart.c
```diff
@@ -450,10 +450,10 @@ static void uart_initPins(void)
 
 /* Device interface */
 
-static ssize_t uart_read(unsigned int minor, addr_t offs, u8 *buff, unsigned int len, unsigned int timeout)
+static ssize_t uart_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
```
agkaminski, line 453: Now there are now an instance of void pointer arithmetic in this function

devices/uart-imxrt117x/uart.c
```diff
@@ -526,9 +526,9 @@ static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
 }
 
 
-static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
+static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const void *buff, size_t len)
```
agkaminski, line 529: Now there are now an instance of void pointer arithmetic in this function

agkaminski (changes requested):
This introduced many instances of void pointer arithmetics. It is undefined behaviour - void pointer arithmetic is an GNU extention

devices/uart-imxrt106x/uart.c
```diff
@@ -511,15 +511,14 @@ static ssize_t uart_write(unsigned int minor, const u8 *buff, unsigned int len)
...
 {
-	unsigned int l;
+	ssize_t res;
+	size_t l;
 
-	for (l = 0; len;) {
-		if ((l = uart_write(minor, buff, len)) < 0)
+	for (l = 0; l < len; l += res) {
+		if ((res = uart_write(minor, (const char *)buff + l, len - l)) < 0)
```
agkaminski, line 520: Comment for the future - it's best not to use lowercase L (l) because it can be mistaken either with 1 or I (uppercase i)

agkaminski: l1I - which is which? 😄

Maxez:
@agkaminski Problems with void * arithmetics are fixed now. I also changed ```uart_safeWrite()``` implementation a bit (please take a look this):
Before:
```c
static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const u8 *buff, unsigned int len)
{
	unsigned int l;

	for (l = 0; len;) {
		if ((l = uart_write(minor, buff, len)) < 0)
			return -ENXIO;
		buff += l;
		len -= l;
	}

	return len;
}
```

and after change:
```c
static ssize_t uart_safeWrite(unsigned int minor, addr_t offs, const void *buff, size_t len)
{
	ssize_t res;
	size_t l;

	for (l = 0; l < len; l += res) {
		if ((res = uart_write(minor, (const char *)buff + l, len - l)) < 0)
			return -ENXIO;
	}

	return len;
}
```

agkaminski (approved):
LGTM

Regarding uart_safeWrite() - also LGTM

Maxez:
> 
> 
> Actually... What if uart_safeWrite() return zero? This case doesn't seem to handled in any way (in both version) - we'll potentialy get infinite loop this way. Not a part of this PR, but something to be fixed in the future PR

I'm not sure what's the problem with returning zero. Where would the infinite loop happen? I think zero is a valid return value.

agkaminski:
> 
> 
> > Actually... What if uart_safeWrite() return zero? This case doesn't seem to handled in any way (in both version) - we'll potentialy get infinite loop this way. Not a part of this PR, but something to be fixed in the future PR
> 
> I'm not sure what's the problem with returning zero. Where would the infinite loop happen? I think zero is a valid return value.

if res == 0, then we won't return (because only res < 0 causes return). l += res == l, so if uart_write should return 0 constantly, then we'll be stuck in a infine loop.

But I quess that uart_write returning 0 constantly shouldn't be really a case... So maybe there's no need to handle such case.

Maxez:
Oh, I thought you meant `uart_safeWrite()` returning 0, not `uart_write()`. You're right - we'd get stuck in an infinite loop. We assume that tx fifo gets consumed. As you mentioned not a part of this PR so I'm leaving it for the future.

syspage.c
```diff
@@ -518,9 +516,7 @@ int syspage_addmap(const char *name, addr_t start, addr_t end, const char *attr)
 	if (syspage_strAttr2ui(attr, &map->attr) < 0)
 		return -EINVAL;
 
-	len = hal_strlen(name);
-	size = min(len, SIZE_MAP_NAME - 1);
-
+	size = min(hal_strlen(name), SIZE_MAP_NAME - 1);
```
HBuczynski, line 519: In your change `hal_strlen` is invoked twice in `min`. It was the main reason why there is added `len`.
Yeah, your change remove one line of code, but I don't think that it is a better approach.

HBuczynski (changes requested):
The `string.h` files have been copied from `phoenix-rtos-kernel`. I think that changes which you added within this PR should be also implemented in `phoenix-rtos-kernel`.

syspage.c
Maxez: Thanks for the explanation. I didn't realize that min macro would force hal_strlen() evaluation twice. I restored the original version.

agkaminski (approved):

Maxez:
> The `string.h` files have been copied from `phoenix-rtos-kernel`. I think that changes which you added within this PR should be also implemented in `phoenix-rtos-kernel`.

I created issue with proposal to change standard functions interface among Phoenix-RTOS code base ([link](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/129)).

agkaminski:
> 
> 
> > The `string.h` files have been copied from `phoenix-rtos-kernel`. I think that changes which you added within this PR should be also implemented in `phoenix-rtos-kernel`.
> 
> I created issue with proposal to change standard functions interface among Phoenix-RTOS code base ([link](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/129)).

Just to clarify - this isn't in the scope of this PR and is not merge blocker.

HBuczynski (approved):

HBuczynski:
> > > The `string.h` files have been copied from `phoenix-rtos-kernel`. I think that changes which you added within this PR should be also implemented in `phoenix-rtos-kernel`.
> > 
> > 
> > I created issue with proposal to change standard functions interface among Phoenix-RTOS code base ([link](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/129)).
> 
> Just to clarify - this isn't in the scope of this PR and is not merge blocker.

Yes, I agree that it isn't a merge blocker and scope of this PR.
However some changes have been done and I think that somewhere should be written that similar work need to be done in `phoenix-rtos-kernel`.

agkaminski:
> 
> 
> > > > The `string.h` files have been copied from `phoenix-rtos-kernel`. I think that changes which you added within this PR should be also implemented in `phoenix-rtos-kernel`.
> > > 
> > > 
> > > I created issue with proposal to change standard functions interface among Phoenix-RTOS code base ([link](https://github.com/phoenix-rtos/phoenix-rtos-project/issues/129)).
> > 
> > 
> > Just to clarify - this isn't in the scope of this PR and is not merge blocker.
> 
> Yes, I agree that it isn't a merge blocker and scope of this PR.
> However some changes have been done and I think that somewhere should be written that similar work need to be done in `phoenix-rtos-kernel`.

string.h interface in kernel was made not compatible with libc functions on purpose, so I think that such change won't be made.

agkaminski:
If these functions interface were to be like analogous to kernel, then this PR is a change in the wrong direction - as kernel string.h is not compatible with libc functions.

HBuczynski:
> If these functions interface were to be like analogous to kernel, then this PR is a change in the wrong direction - as kernel string.h is not compatible with libc functions.

Yeah, I think that these functions should be analogous to kernel. I think that in the future we will aim to integrate some code between loader and kernel.
If these changes are not accepted in kernel, they should be removed from this PR.

HBuczynski (approved):

Maxez:
@agkaminski Paweł agreed to proposed HAL string function changes (in plo for now and in kernel in the future). I'd like to merge this PR now so please let me know if you also approve it.

agkaminski (approved):

nalajcie:
If You would write something like: 

Fixes phoenix-rtos/phoenix-rtos-project#129

In Your commit msg, the GH issue would be automatically resolved when this PR would be merged

PR #143, opened 2021-10-14
flash-zynq7000: add nor flash driver

HBuczynski:
## Description

The following changes have been made:
 - Quad-SPi controller driver implementation,
 - getting flash configuration based on CFI (Common Flash Interface) which is get via RDID command,
 - nor flash driver implementaion  using qspi.

JIRA: PD-200
## Motivation and Context

Needed for booting zynq7000 from nor flash.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (zynq7000).

## Checklist:

## Special treatment

### DIFF
--- a/devices/flash-zynq7000/flashcfg.c
+++ b/devices/flash-zynq7000/flashcfg.c
@@ -1,0 +1,104 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system loader
+ *
+ * Common Flash Interface for flash drive
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+
+#include "flashcfg.h"
+#include <lib/errno.h>
+
+
+#define MANUFACTURER_ID_SPANSION 0x1
+#define SPANSION_MEM256          0x19
+#define SPANSION_MEM128          0x18
+
+
+static void flashcfg_spansionCmds(flash_info_t *info)
+{
+	info->cmds[flash_cmd_rdid].opCode = FLASH_CMD_RDID;
+	info->cmds[flash_cmd_rdid].size = 1;
+	info->cmds[flash_cmd_rdid].dummyCyc = 24;
+
+	info->cmds[flash_cmd_rdsr1].opCode = FLASH_CMD_RDSR1;
+	info->cmds[flash_cmd_rdsr1].size = 1;
+	info->cmds[flash_cmd_rdsr1].dummyCyc = 8;
+
+	info->cmds[flash_cmd_wrdi].opCode = FLASH_CMD_WRDI;
+	info->cmds[flash_cmd_wrdi].size = 1;
+	info->cmds[flash_cmd_wrdi].dummyCyc = 0;
+
+	info->cmds[flash_cmd_wren].opCode = FLASH_CMD_WREN;
+	info->cmds[flash_cmd_wren].size = 1;
+	info->cmds[flash_cmd_wren].dummyCyc = 0;
+
+	info->cmds[flash_cmd_read].opCode = FLASH_CMD_READ;
+	info->cmds[flash_cmd_read].size = 4;
+	info->cmds[flash_cmd_read].dummyCyc = 0;
+
+	info->cmds[flash_cmd_fast_read].opCode = FLASH_CMD_FAST_READ;
+	info->cmds[flash_cmd_fast_read].size = 4;
+	info->cmds[flash_cmd_fast_read].dummyCyc = 8;
+
+	info->cmds[flash_cmd_dor].opCode = FLASH_CMD_DOR;
+	info->cmds[flash_cmd_dor].size = 4;
+	info->cmds[flash_cmd_dor].dummyCyc = 8;
+
+	info->cmds[flash_cmd_qor].opCode = FLASH_CMD_QOR;
+	info->cmds[flash_cmd_qor].size = 4;
+	info->cmds[flash_cmd_qor].dummyCyc = 8;
+
+	info->cmds[flash_cmd_dior].opCode = FLASH_CMD_DIOR;
+	info->cmds[flash_cmd_dior].size = 4;
+	info->cmds[flash_cmd_dior].dummyCyc = 8;
+
+	info->cmds[flash_cmd_qior].opCode = FLASH_CMD_QIOR;
+	info->cmds[flash_cmd_qior].size = 4;
+	info->cmds[flash_cmd_qior].dummyCyc = 24;
+
+	info->cmds[flash_cmd_pp].opCode = FLASH_CMD_PP;
+	info->cmds[flash_cmd_pp].size = 4;
+	info->cmds[flash_cmd_pp].dummyCyc = 0;
+
+	info->cmds[flash_cmd_qpp].opCode = FLASH_CMD_QPP;
+	info->cmds[flash_cmd_qpp].size = 4;
+	info->cmds[flash_cmd_qpp].dummyCyc = 0;
+
+	info->cmds[flash_cmd_p4e].opCode = FLASH_CMD_P4E;
+	info->cmds[flash_cmd_p4e].size = 4;
+	info->cmds[flash_cmd_p4e].dummyCyc = 0;
+
+	info->cmds[flash_cmd_p64e].opCode = FLASH_CMD_SE;
+	info->cmds[flash_cmd_p64e].size = 4;
+	info->cmds[flash_cmd_p64e].dummyCyc = 0;
+
+	info->cmds[flash_cmd_be].opCode = FLASH_CMD_BE;
+	info->cmds[flash_cmd_be].size = 1;
+	info->cmds[flash_cmd_be].dummyCyc = 0;
+}
+
+
+int flashcfg_infoResolve(flash_info_t *info)
+{
+	int res = EOK;
+
+	if (info->cfi.vendorData[0] == MANUFACTURER_ID_SPANSION && info->cfi.vendorData[2] == SPANSION_MEM256) {
+		info->name = "Spansion s25fl256s1";
+		flashcfg_spansionCmds(info);
+	}
+	else {
+		info->name = "Unknown";
+		res = -EINVAL;
+	}
+
+	return res;
+}
--- a/devices/flash-zynq7000/flashdrv.c
+++ b/devices/flash-zynq7000/flashdrv.c
@@ -1,0 +1,455 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system loader
+ *
+ * Zynq-7000 nor flash driver
+ *
+ * Copyright 2020 Phoenix Systems
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "qspi.h"
+#include "flashcfg.h"
+
+#include <lib/lib.h>
+#include <devices/devs.h>
+
+
+#define TIMEOUT_CMD_MS 0x05
+#define MAX_SIZE_CMD   0x20
+
+
+struct {
+	/* Buffers for command transactions */
+	u8 cmdRx[MAX_SIZE_CMD];
+	u8 cmdTx[MAX_SIZE_CMD];
+
+	/* Data caching in sector's buffer */
+	u8 *buff;
+	u32 buffPos;
+	u32 regID;
+	u32 sectID;
+
+	flash_info_t info;
+} fdrv_common;
+
+
+/* Auxiliary flash commands */
+
+static int flashdrv_cfiRead(flash_cfi_t *cfi)
+{
+	ssize_t res;
+	size_t cmdSz, dataSz;
+	/* RDID instruction is a generic for each flash memory */
+	const flash_cmd_t cmds = { .opCode = FLASH_CMD_RDID, .size = 1, .dummyCyc = 24 };
+
+	/* Cmd size is rounded to 4 bytes, it includes dummy cycles [b]. */
+	cmdSz = (cmds.size + cmds.dummyCyc / 8 + 0x3) & ~0x3;
+	/* Size of the data which is received during cmd transfer */
+	dataSz = cmds.dummyCyc / 8;
+
+	hal_memset(fdrv_common.cmdTx, 0, cmdSz);
+	fdrv_common.cmdTx[0] = cmds.opCode;
+
+	qspi_start();
+	if ((res = qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmdSz, TIMEOUT_CMD_MS)) < 0) {
+		qspi_stop();
+		return res;
+	}
+
+	/* Copy data received after dummy bytes */
+	hal_memcpy(cfi, (fdrv_common.cmdRx + cmds.size), dataSz);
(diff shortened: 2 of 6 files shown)


devices/flash-zynq7000/flashcfg.c
```diff
@@ -0,0 +1,104 @@
...
+#define MANUFACTURER_ID_SPANSION 0x1
+#define SPANSION_MEM256          0x19
+#define SPANSION_MEM128          0x18
+
+
+static void flashcfg_spansionCmds(flash_info_t *info)
+{
+	info->cmds[flash_cmd_rdid].opCode = FLASH_CMD_RDID;
+	info->cmds[flash_cmd_rdid].size = 1;
+	info->cmds[flash_cmd_rdid].dummyCyc = 24;
```
agkaminski, line 30: It would be best to just place it in .rodata as static const table, but I don't see way to do this and keep readability (i.e. tying enum and define in clear way), so I propose this middle-ground: 
```suggestion
	info->cmds[flash_cmd_rdid] = { FLASH_CMD_RDID, 1, 24 };
```

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+	hal_memset(fdrv_common.cmdTx, 0, cmdSz);
+	fdrv_common.cmdTx[0] = cmds.opCode;
+
+	qspi_start();
+	if ((res = qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmdSz, TIMEOUT_CMD_MS)) < 0)
+		return res;
+
+	/* Copy data received after dummy bytes */
+	hal_memcpy((u8 *)cfi, (const void *)(fdrv_common.cmdRx + cmds.size), dataSz);
```
agkaminski, line 64: Is this cast to `const void *` necessary?

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+	if ((res = qspi_polledTransfer(NULL, (u8 *)cfi + dataSz, sizeof(flash_cfi_t) - dataSz, 5 * TIMEOUT_CMD_MS)) < 0) {
+		qspi_stop();
+		return res;
+	}
+	qspi_stop();
+
+	return EOK;
```
agkaminski, line 72:
```suggestion
	res = qspi_polledTransfer(NULL, (u8 *)cfi + dataSz, sizeof(flash_cfi_t) - dataSz, 5 * TIMEOUT_CMD_MS);
	qspi_stop();
	return res;
```

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+
+static int flashdrv_statusRegGet(unsigned int cmdID, unsigned int *val)
+{
+	ssize_t res;
+	size_t cmdSz;
+	const flash_cmd_t *cmd;
+
+	if (cmdID != flash_cmd_rdsr1)
+		return -EINVAL;
```
agkaminski, line 83: Why do we even pass `cmdID` to this function if only one is acceptable?

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+	*val = 0;
+	cmd = &fdrv_common.info.cmds[cmdID];
+
+	cmdSz = cmd->size + cmd->dummyCyc / 8;
+	hal_memset(fdrv_common.cmdTx, 0, cmdSz);
+	fdrv_common.cmdTx[0] = cmd->opCode;
+
+	qspi_start();
+	if ((res = qspi_polledTransfer(fdrv_common.cmdTx, (u8 *)val, cmdSz, TIMEOUT_CMD_MS)) < 0)
+		return res;
```
agkaminski, line 94: Oops, no `qspi_stop()` on this path

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+		return -EINVAL;
+
+	cmd = &fdrv_common.info.cmds[cmdID];
+
+	hal_memset(fdrv_common.cmdTx, 0, cmd->size);
+	fdrv_common.cmdTx[0] = cmd->opCode;
+
+	qspi_start();
+	if ((res = qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmd->size, TIMEOUT_CMD_MS)) < 0)
+		return res;
```
agkaminski, line 116: Oops, no `qspi_stop()` on this path

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+	if ((i = flashdrv_regionFind(offs)) < 0)
+		return i;
+
+	switch (CFI_SIZE_SECTION(cfi->regs[i].size)) {
+		case 0x1000:
+			cmd = &fdrv_common.info.cmds[flash_cmd_p4e];
+			break;
+		case 0x10000:
+			cmd = &fdrv_common.info.cmds[flash_cmd_se];
+			break;
```
agkaminski, line 163: This whole thing looks like it's our Spantion memory specific. It might stay, but maybe there is some way to make this more generic?

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+	hal_memset(fdrv_common.cmdTx, 0, cmd->size);
+	fdrv_common.cmdTx[0] = cmd->opCode;
+	fdrv_common.cmdTx[1] = (offs >> 16) & 0xff;
+	fdrv_common.cmdTx[2] = (offs >> 8) & 0xff;
+	fdrv_common.cmdTx[3] = offs & 0xff;
+
+	qspi_start();
+	if ((res = qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmd->size, TIMEOUT_CMD_MS)) < 0)
+		return res;
```
agkaminski, line 179: Oops, no `qspi_stop()` on this path

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+	timeout = CFI_TIMEOUT_MAX_PROGRAM(cfi->timeoutTypical.pageWrite, cfi->timeoutMax.pageWrite) + TIMEOUT_CMD_MS;
+	start = hal_timerGet();
+	do {
+		if ((res = flashdrv_statusRegGet(flash_cmd_rdsr1, &st)) < 0)
+			return res;
+
+		if ((hal_timerGet() - start) >= timeout)
+			return -ETIME;
+	} while ((st >> 8) & 0x1);
```
agkaminski, line 244: Copy-paste code, making this a function might be a good idea

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+	if (fdrv_common.regID)
+		regStart = CFI_SIZE_REGION(cfi->regs[fdrv_common.regID - 1].size, cfi->regs[fdrv_common.regID - 1].count);
+
+	sectSz = CFI_SIZE_SECTION(cfi->regs[fdrv_common.regID].size);
+	pageSz = CFI_SIZE_PAGE(cfi->pageSize);
+	pgNb = sectSz / pageSz;
+
+	for (i = 0; i < pgNb; ++i) {
+		dst = regStart + fdrv_common.sectID * sectSz + i * pageSz;
+		src = (const u8 *)(fdrv_common.buff + i * pageSz);
```
agkaminski, line 316: Is this cast necessary?

devices/flash-zynq7000/flashdrv.c
```diff
@@ -0,0 +1,461 @@
...
+
+		regStart = regID ? CFI_SIZE_REGION(cfi->regs[regID - 1].size, cfi->regs[regID - 1].count) : 0;
+		sectSz = CFI_SIZE_SECTION(cfi->regs[regID].size);
+		sectID = (offs - regStart) / sectSz;
+
+		if (regID != fdrv_common.regID || sectID != fdrv_common.sectID) {
+			if ((res = flashdrv_sync(minor)) < 0)
+				return res;
+
+			timeout = sectSz / 0x100 * TIMEOUT_CMD_MS;
```
agkaminski, line 361: Where does this 0x100 come from?

agkaminski: Also, make operation order explicit, it does matter when operating on integers

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+			*(qspi_common.base + txd01) = buff[0];
+			return 1;
+
+		case 2:
+			*(qspi_common.base + txd10) = buff[0] | (buff[0] >> 8);
+			return 2;
+
+		case 3:
+			*(qspi_common.base + txd11) = buff[0] | (buff[0] >> 8) | (buff[0] >> 16);
+			return 3;
```
kemonats, line 108: Is this code correct?
`buf[0]` is of type `u8` and shift it to the right by 8 or 16 gives `0`.

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+			default:
+				rxBuff[0] = data & 0xff;
+				rxBuff[1] = (data >> 8) & 0xff;
+				rxBuff[2] = (data >> 16) & 0xff;
+				rxBuff[3] = (data >> 24) & 0xff;
+				return 4;
+		}
+	}
+
+	return (size < 4) ? size : 4;
```
agkaminski, line 85: Dead code, also it will return size, when `rxBuff` was NULL (in what case it should return error or zero). Change returns to breaks in switch-case and add case for rxBuff == NULL

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+}
+
+
+static unsigned int qspi_txData(const u8 *txBuff, size_t size)
+{
+	const u32 dummy = 0;
+	const u8 *buff = (txBuff == NULL) ? (const u8 *)&dummy : txBuff;
```
agkaminski, line 92:
```suggestion
	const u8 dummy[4] = { 0 };
	const u8 *buff = (txBuff == NULL) ? dummy : txBuff;
```

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+			*(qspi_common.base + txd10) = buff[0] | (buff[0] >> 8);
+			return 2;
+
+		case 3:
+			*(qspi_common.base + txd11) = buff[0] | (buff[0] >> 8) | (buff[0] >> 16);
+			return 3;
+
+		case 4:
+		default:
+			*(qspi_common.base + txd00) = *((u32 *)buff);
```
agkaminski, line 112: This might be unaligned access. Use intermediate variable and memcpy instead

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+				break;
+
+			tempSz = qspi_rxData(rxBuff, rxSz);
+
+			rxSz -= tempSz;
+			rxBuff += tempSz;
+		}
+
+		if ((hal_timerGet() - start) >= timeout)
+			return -ETIME;
```
agkaminski, line 167: Do we need timeout here?

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+		case mio_pin_02:
+		case mio_pin_03:
+		case mio_pin_04:
+		case mio_pin_05:
+		case mio_pin_06:
+			break;
+
+		default:
+			return -EINVAL;
+	}
```
agkaminski, line 203:
```suggestion
	if (pin == mio_pin_01) {
		ctl.pullup = 1;
		ctl.speed = 0x0;
	}
	else if (pin < mio_pin_01 &&  pin > mio_pin_06) {
		return -EINVAL;
	}
```

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+}
+#endif
+
+
+static void qspi_IOMode(void)
+{
+	/* Configure I/O mode */
+
+	/* Disable QSPI */
+	*(qspi_common.base + er) &= ~0x1;
```
agkaminski, line 269: Add memory barier after this one

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+	*(qspi_common.base + cr) |= (0x3 << 6);
+	/* Set clock phase and polarity */
+	*(qspi_common.base + cr) &= ~(0x3 << 1);
+
+
+	/* Enable manual mode and manual CS */
+	*(qspi_common.base + cr) |= (0x3 << 14);
+
+	/* Disable linear mode */
+	*(qspi_common.base + lqspi_cr) = 0;
```
agkaminski, line 290: And memory barier after this one. Maybe some are needed in between (I think not)?

devices/flash-zynq7000/qspi.c
```diff
@@ -0,0 +1,329 @@
...
+	*(qspi_common.base + cr) |= (1 << 10);
+	*(qspi_common.base + er) = 0;
+}
+
+
+void qspi_start(void)
+{
+	*(qspi_common.base + rxth) = 0x1;
+	*(qspi_common.base + cr) &= ~(1 << 10);
+	*(qspi_common.base + er) = 0x1;
```
agkaminski, line 47: Add memory barrier after this one

devices/flash-zynq7000/flashdrv.c
HBuczynski: Done

devices/flash-zynq7000/flashdrv.c
HBuczynski: Done

devices/flash-zynq7000/flashdrv.c
HBuczynski: Initial approach allowed to get values from different status registers. However in the final code only `RDSR1` is used. I agree with you that these function doesn't need `cmdID`.
Done.

devices/flash-zynq7000/flashdrv.c
HBuczynski: Done

devices/flash-zynq7000/flashdrv.c
HBuczynski: Nope. Removed.

devices/flash-zynq7000/qspi.c
HBuczynski: Nope this code is not correct. Thank You for noting it!!

It's just an ugly copy-paste from the other function (from the code developing), which cause that I have left it as it is. While I have tested the code I usually have sent data to program with size `0x100` therefore it have't gone to these cases.
When I have sent status commands, it has went to these cases but sent the dummy data so it looked ok. 
To sump up,  it was critical error, so thanks again for pointing it out. 

Fixed.

devices/flash-zynq7000/flashdrv.c
HBuczynski: done

devices/flash-zynq7000/flashdrv.c
HBuczynski: Yeah, good suggestion. Moved this code to separate function.

devices/flash-zynq7000/flashdrv.c
HBuczynski: Nope. removed

devices/flash-zynq7000/qspi.c
HBuczynski: Done

PR #167, opened 2022-03-16
Add linker scripts tailored for plo

gerard5:
JIRA: NIL-181

The main purpose of this commit is to allow critical sections of code to run from `RAM` instead of `flash`.

An example here is the `i.MX RT` CPU series on which `FlexSPI` is configured (without using the `ROM-API` and where it is not supported), the part of code launched from flash (`XIP`) at the time of `FlexSPI` bus reconfiguration must be executed from the RAM, so this part of plo and IRQ handlers are reloacated to RAM, the rest may reside on flash.

**This change entails the addition of linker scripts for all supported by plo platforms as well, so that the way of plo building is uniform.**

This commit adds support for the following sections on all platforms supported by `plo`:
* `.init` (where bootstrap code is placed, like boot-sector)
* `.fastram.text.rel` (aliased with `.noxip` section)
* `.data`
* `.heap` (syspage, and dynamic allocation)
* `.stack`

The configuration of platform and memory layout is driven by `linker script & C template` system located in `ld` subdirectory and main `Makefile`.

TODO: `IA-32` linker script is to be enhanced in near future with `.gdt`, and `.idt` descriptor sub-sections located in `.data` section.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (evk-mimxrt1064, evk-mimxrt1176, stm32l4x6, qemu-ia32-generic, qemu-zynq,).

## Checklist:

## Special treatment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 #
 # Makefile for plo (structured)
 #
-# Copyright 2020 Phoenix Systems
+# Copyright 2020-2022 Phoenix Systems
 #
 # %LICENSE%
 #
@@ -28,6 +28,9 @@ include cmds/Makefile
 OBJS += $(addprefix $(PREFIX_O), _startc.o plo.o syspage.o)
 
 
+.PHONY: all base ram clean
+
+
 all: base ram
 
 
@@ -37,51 +40,55 @@ base: $(PREFIX_PROG_STRIPPED)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf  $(PRE
 ram: $(PREFIX_PROG_STRIPPED)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf $(PREFIX_PROG_STRIPPED)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).img
 
 
-$(BUILD_DIR)/script.plo:
-	@printf "TOUCH script.plo\n"
-	$(SIL)touch $(BUILD_DIR)/script.plo
+$(BUILD_DIR)/script.plo $(BUILD_DIR)/ramscript.plo:
+	@echo "TOUCH $(@F)"
+	$(SIL)touch $@
 
 
 $(PREFIX_O)/script.o.plo: $(PREFIX_O)cmds/cmd.o $(BUILD_DIR)/script.plo
 	@mkdir -p $(@D)
-	@printf "EMBED script.plo\n"
-	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/script.plo $(PREFIX_O)cmds/cmd.o --add-symbol script=.data:0 $(PREFIX_O)script.o.plo
-
-
-$(BUILD_DIR)/ramscript.plo:
-	@printf "TOUCH script.plo\n"
-	$(SIL)touch $(BUILD_DIR)/ramscript.plo
+	@echo "EMBED script.plo"
+	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/script.plo $(PREFIX_O)cmds/cmd.o --add-symbol script=.data:0 $@
 
 
 $(PREFIX_O)/ramscript.o.plo: $(PREFIX_O)cmds/cmd.o $(BUILD_DIR)/ramscript.plo
 	@mkdir -p $(@D)
-	@printf "EMBED ramscript.plo\n"
-	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/ramscript.plo $(PREFIX_O)cmds/cmd.o --add-symbol script=.data:0 $(PREFIX_O)ramscript.o.plo
+	@echo "EMBED ramscript.plo"
+	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/ramscript.plo $(PREFIX_O)cmds/cmd.o --add-symbol script=.data:0 $@
 
 
-$(PREFIX_PROG)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(OBJS) $(PREFIX_O)/script.o.plo
+$(PREFIX_PROG)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld $(OBJS) $(PREFIX_O)/script.o.plo
 	@mkdir -p $(@D)
-	@(printf "LD  %-24s\n" "$(@F)");
-	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(INIT_FLASH) $(BSS) -o $(PREFIX_PROG)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf $(OBJS) $(PREFIX_O)/script.o.plo $(GCCLIB)
+	@echo "LD  $(@F)"
+	$(SIL)$(LD) $(LDFLAGS) -Map=$<.map -o $@ -T $^ $(GCCLIB)
 
 
-$(PREFIX_PROG)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(OBJS) $(PREFIX_O)/ramscript.o.plo
+$(PREFIX_PROG)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)-ram.ld $(OBJS) $(PREFIX_O)/ramscript.o.plo
 	@mkdir -p $(@D)
-	@(printf "LD  %-24s\n" "$(@F)");
-	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(INIT_RAM) $(BSS) -o $(PREFIX_PROG)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf $(OBJS) $(PREFIX_O)/ramscript.o.plo $(GCCLIB)
+	@echo "LD  $(@F)"
+	$(SIL)$(LD) $(LDFLAGS) -Map=$<.map -o $@ -T $^ $(GCCLIB)
 
 
 $(PREFIX_PROG_STRIPPED)%.hex: $(PREFIX_PROG_STRIPPED)%.elf
-	@(printf "HEX %s\n" "$(@F)");
+	@echo "HEX $(@F)"
 	$(SIL)$(OBJCOPY) -O ihex $< $@
 
 
 $(PREFIX_PROG_STRIPPED)%.img: $(PREFIX_PROG_STRIPPED)%.elf
-	@(printf "BIN %s\n" "$(@F)");
+	@echo "BIN $(@F)"
 	$(SIL)$(OBJCOPY) -O binary $< $@
 
 
-.PHONY: clean
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld:
+	@echo "GEN $(@F)"
+	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+
+
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)-%.ld:
+	@echo "GEN $(@F)"
+	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -D$* -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+
+
 clean:
 	@echo "rm -rf $(BUILD_DIR)"
 
--- a/hal/armv7a/imx6ull/hal.c
+++ b/hal/armv7a/imx6ull/hal.c
@@ -5,8 +5,8 @@
  *
  * Hardware Abstraction Layer
  *
- * Copyright 2021 Phoenix Systems
- * Author: Hubert Buczynski
+ * Copyright 2021-2022 Phoenix Systems
+ * Author: Hubert Buczynski, Gerard Swiderski
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -24,8 +24,11 @@ struct {
 
 
 /* Linker symbols */
-extern void _end(void);
-extern void _plo_bss(void);
+extern char __ramtext_start[], __ramtext_end[];
+extern char __data_start[], __data_end[];
+extern char __bss_start[], __bss_end[];
+extern char __heap_base[], __heap_limit[];
+extern char __stack_top[], __stack_limit[];
 
 /* Timer */
 extern void timer_init(void);
@@ -134,8 +137,16 @@ static void hal_getMinOverlappedRange(addr_t start, addr_t end, mapent_t *entry,
 
 int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
 {
-	mapent_t tempEntry;
-	mapent_t minEntry;
+	int i;
+	mapent_t tempEntry, minEntry;
+
+	static const mapent_t entries[] = {
+		{ .start = (addr_t)__ramtext_start, .end = (addr_t)__ramtext_end, .type = hal_entryTemp },
+		{ .start = (addr_t)__data_start, .end = (addr_t)__data_end, .type = hal_entryTemp },
+		{ .start = (addr_t)__bss_start, .end = (addr_t)__bss_end, .type = hal_entryTemp },
+		{ .start = (addr_t)__heap_base, .end = (addr_t)__heap_limit, .type = hal_entryTemp },
+		{ .start = (addr_t)__stack_limit, .end = (addr_t)__stack_top, .type = hal_entryTemp },
+	};
 
 	if (start == end)
 		return -1;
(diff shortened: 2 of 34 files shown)


hal/armv7m/imxrt/117x/config.h
```diff
@@ -30,17 +30,13 @@
...
-#endif
-
-
 #define PATH_KERNEL "phoenix-armv7m7-imxrt117x.elf"
 #define CPU_INFO    "Cortex-M i.MX RT117x"
 
-#define SIZE_PAGE    0x200
-#define SIZE_STACK   (8 * SIZE_PAGE)
-#define SIZE_SYSPAGE (8 * SIZE_PAGE)
+#endif
```
agkaminski, line 36: Why move this endif?

ld/armv7a7-imx6ull.ldt
```diff
@@ -0,0 +1,54 @@
...
+#define ADDR_OCRAM 0x907000
+#define SIZE_OCRAM 0x020000
+
+
+#if defined(__LINKER__)
+
+/* Memory map setup */
+MEMORY
+{
+	m_ocram   (rwx) : ORIGIN = 0x00907000, LENGTH = 128k
```
agkaminski, line 38: Not really - we only have 68K usable, lower adresses are reserved by the bootloader

agkaminski: Usable of course if we take into the account we're starting from 0x00907000 - after bootloader completes memory below this address become available again

ld/armv7m4-stm32l4x6.ldt
```diff
@@ -0,0 +1,51 @@
...
+{
+	m_sram2   (rwx) : ORIGIN = 0x20040000, LENGTH = 256k
+	m_sram1   (rwx) : ORIGIN = 0x20010000, LENGTH = 192k
+	m_flash2  (rx)  : ORIGIN = 0x08000000, LENGTH = 512k
+	m_flash1  (rx)  : ORIGIN = 0x00000000, LENGTH = 512k
+}
+
+/* FLASH image */
+REGION_ALIAS("PLO_IMAGE", m_flash2);
+REGION_ALIAS("TCM_TEXT", m_flash2);
```
agkaminski, line 40: Everything resides inside of flash1 - flash2 is a second banks, reserved for OTA purpose

ld/armv7m7-imxrt117x.ldt
```diff
@@ -0,0 +1,63 @@
...
+
+#if defined(__LINKER__)
+
+/* Memory map setup */
+MEMORY
+{
+	m_itcm    (rwx) : ORIGIN = 0x00000000, LENGTH = 8 * 32k
+	m_itext   (rwx) : ORIGIN = 0x0003f000, LENGTH = 4k
+	m_dtcm    (rw)  : ORIGIN = 0x20002000, LENGTH = 5 * 32k
+	m_ocram   (rwx) : ORIGIN = 0x20250000, LENGTH = 5 * 32k
```
agkaminski, line 34: imxrt117x have two ocrams - I don't see them here...

agkaminski (commented):
Before merging I would like confirmation (from different people, no problem) that it's working as intended on all supported targets.

ld/armv7m7-imxrt106x.ldt
```diff
@@ -0,0 +1,66 @@
...
+/* FlexRAM configuration */
+FLEXRAM_CONFIG = 0xaaaaabff;
+
+/* Memory map setup */
+MEMORY
+{
+	m_itcm    (rwx) : ORIGIN = 0x00000000, LENGTH = 5 * 32k
+	m_itext   (rwx) : ORIGIN = 0x00027000, LENGTH = 4 * 1k
+	m_dtcm    (rw)  : ORIGIN = 0x20002000, LENGTH = 11 * 32k
+	m_ocram   (rwx) : ORIGIN = 0x20200000, LENGTH = 0 * 32k
```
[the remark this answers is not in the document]
gerard5: Agree, it is doable.

ld/armv7m4-stm32l4x6.ldt
```diff
@@ -0,0 +1,51 @@
...
+REGION_ALIAS("DATA", m_sram1);
+REGION_ALIAS("BSS", m_sram1);
+REGION_ALIAS("HEAP", m_sram1);
+REGION_ALIAS("STACK", m_sram1);
```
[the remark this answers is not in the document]
gerard5: Originally (currently) `stack_top` resides there https://github.com/phoenix-rtos/plo/blob/cd80102775f353e995977aa3219657c4cfad4a3f/hal/armv7m/stm32/l4/config.h#L41 I placed it there too (in this PR), You may easily verify with `readelf`:
```
nth paddr          size vaddr         vsize perm name                                                                                                                                                                                   
―――――――――――――――――――――――――――――――――――――――――――――――――――――
...
7   0x000089e0     0xdc 0x20010000     0xdc -rw- .data
8   0x00008ac0      0x0 0x200100e0   0x132c -rw- .bss
9   0x00008c00      0x0 0x20011600   0x1000 -rw- .heap
10  0x00008ac0      0x0 0x20012600   0x1000 -rw- .stack
...
```

So it is like before but now stack has it's own section (and `.data` is in sram, not on flash as it was previously). I can place stack on `stm32l4x6` at the end of ram if You wish, but I followed the (before linker scripts) current approach for each platform. Now it is easy to place `stack_top` at 384kB, if it is desired, I may put it there.

ld/armv7m4-stm32l4x6.ldt
```diff
@@ -0,0 +1,51 @@
...
+
+#if defined(__LINKER__)
+
+/* Memory map setup */
+MEMORY
+{
+	m_sram2   (rwx) : ORIGIN = 0x20040000, LENGTH = 256k
+	m_sram1   (rwx) : ORIGIN = 0x20010000, LENGTH = 192k
+	m_flash2  (rx)  : ORIGIN = 0x08000000, LENGTH = 512k
+	m_flash1  (rx)  : ORIGIN = 0x00000000, LENGTH = 512k
```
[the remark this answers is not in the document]
gerard5: Thanks, changed origins to: `flash1=0x08000000`, `flash2=0x08080000`.

ld/armv7m4-stm32l4x6.ldt
gerard5: Thanks, changed to `flash1`

hal/armv7m/imxrt/117x/config.h
gerard5: Thanks, restored back.

ld/armv7m7-imxrt117x.ldt
gerard5: Thanks, added now.

ld/armv7m7-imxrt117x.ldt
```diff
@@ -0,0 +1,63 @@
...
+#define SIZE_HEAP  (8 * SIZE_PAGE)
+
+#if defined(__LINKER__)
+
+/* Memory map setup */
+MEMORY
+{
+	m_itcm    (rwx) : ORIGIN = 0x00000000, LENGTH = 8 * 32k
+	m_itext   (rwx) : ORIGIN = 0x0003f000, LENGTH = 4k
+	m_dtcm    (rw)  : ORIGIN = 0x20002000, LENGTH = 5 * 32k
```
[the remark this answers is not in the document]
gerard5: Thx, set to 8*32KB.

ld/armv7a7-imx6ull.ldt
gerard5: Reused setup from there https://github.com/phoenix-rtos/plo/blob/cd80102775f353e995977aa3219657c4cfad4a3f/hal/armv7a/imx6ull/config.h#L44-L45

Makefile
```diff
@@ -37,51 +40,55 @@ base: $(PREFIX_PROG_STRIPPED)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf  $(PRE
 ram: $(PREFIX_PROG_STRIPPED)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf $(PREFIX_PROG_STRIPPED)plo-ram-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).img
 
 
-$(BUILD_DIR)/script.plo:
-	@printf "TOUCH script.plo\n"
-	$(SIL)touch $(BUILD_DIR)/script.plo
+$(BUILD_DIR)/script.plo $(BUILD_DIR)/ramscript.plo:
+	@echo "TOUCH $(@F)\n"
```
kemonats, line 44:
```suggestion
	@echo "TOUCH $(@F)"
```

gerard5: Thanks.

ld/armv7m4-stm32l4x6.ldt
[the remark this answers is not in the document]
agkaminski: It's ok as is, right

HBuczynski (approved):
LGTM, 
I have tested Your PR on zynq-7000, imx6ull and ia32-generic and it works fine.

agkaminski (dismissed):
LGTM

Maxez:
@gerard5 we shouldn't initialize structs in `plo` with ` = { 0 }` since the compiler may generate `memset()` for such initialization (which we don't have). Please restore previous `hal_memset()` calls for structs initialization.

gerard5:
> @gerard5 we shouldn't initialize structs in `plo` with ` = { 0 }` since the compiler may generate `memset()` for such initialization (which we don't have). Please restore previous `hal_memset()` calls for structs initialization.

In that `= { 0 }` case it's initialized from `.bss` section, if stucture is initialized eg. `= { .start = (addr_t) -1, 0 }` it's set from `.data` section which is now supported. If data is used as readonly it goes from `.rodata`. No `memset` and `memcpy` is used after `hal_memcpy` and `hal_memset` do it's job in `_startc(argc, argv)` function.

hal/armv7a/imx6ull/hal.c
```diff
@@ -134,33 +137,33 @@ static void hal_getMinOverlappedRange(addr_t start, addr_t end, mapent_t *entry,
 
 int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
 {
-	mapent_t tempEntry;
-	mapent_t minEntry;
+	int i;
+	mapent_t tempEntry = { 0 };
```
agkaminski, line 141: It's not .bss

gerard5: Ok, changing it to draft - as not ready to be merged.

nalajcie:
Just a general comment, there is no safe way to make `gcc` toolchain not insert `mem*()` functions (the `-ffreestanding`, `-fno-builtin` or `-nostdlib` doesn't prevent that) - which can be triggered also by other code structures than const initialization.

Alas, this is clearly stated on gcc website as a requirement (https://gcc.gnu.org/onlinedocs/gcc/Standards.html):
image

So we **need** to provide these functions in our freestanding environments for `gcc` to work correctly (either by rename, wrapper or symbol alias). This is out of scope of this PR, so should be taken care of independently.

hal/armv7a/imx6ull/hal.c
gerard5: In that case C used it's assignment (based on selected optimization) for `ia32` it's the worst case as I see:
2022-03-18-161619_1638x383_scrot
On ARM it uses just two stores.

agkaminski (dismissed):
LGTM

Maxez (approved):

Makefile
```diff
@@ -37,51 +40,55 @@ base: $(PREFIX_PROG_STRIPPED)plo-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf  $(PRE
...
-.PHONY: clean
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld:
+	@echo "GEN $(@F)"
+	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+
+
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)-%.ld:
+	@echo "GEN $(@F)"
+	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -D$* -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+
```
kemonats, line 90:
```suggestion
LD_FILES := ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt
ifeq ($(findstring  arm,$(TARGET_FAMILY)),arm)
  LD_FILES += ld/common/plo-arm.lds
else ifeq ($(findstring  ia32,$(TARGET_FAMILY)),ia32)
  LD_FILES += ld/common/plo-ia32.lds
endif


$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld: $(LD_FILES)
	@echo "GEN $(@F)"
	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -undef -xc -E -P $< > $@


$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)-%.ld: $(LD_FILES)
	@echo "GEN $(@F)"
	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -D$* -undef -xc -E -P $< > $@

```
Changing the `*.ldt` or `*.lds` file will create a `*.ld` file again.

ld/armv7a7-imx6ull.ldt
```diff
@@ -0,0 +1,54 @@
...
+/* Platform specific definitions */
+#define SIZE_PAGE    0x1000
+#define SIZE_STACK   (2 * SIZE_PAGE)
+#define SIZE_HEAP    (2 * SIZE_PAGE)
+
+#define ADDR_DDR 0x80000000
+#define SIZE_DDR (128 * 1024 * 1024)
+
+#define ADDR_OCRAM 0x907000
+#define SIZE_OCRAM 0x11000
```
kemonats, line 30: This value is probably to less.
Linker error:
```
LD  plo-armv7a7-imx6ull.elf
LD  plo-ram-armv7a7-imx6ull.elf
arm-phoenix-ld: /tmp/plo-test/_build/armv7a7-imx6ull/prog/plo-ram-armv7a7-imx6ull.elf section `.bss' will not fit in region `m_ocram'
arm-phoenix-ld: /tmp/plo-test/_build/armv7a7-imx6ull/prog/plo-armv7a7-imx6ull.elf section `.bss' will not fit in region `m_ocram'
arm-phoenix-ld: region `m_ocram' overflowed by 36864 bytes
arm-phoenix-ld: region `m_ocram' overflowed by 36864 bytes
```
Curently it is [link](https://github.com/phoenix-rtos/plo/blob/76b9b92d6adbf734882ffbf7cd7b37e573006720/hal/armv7a/imx6ull/config.h#L45)
```
#define SIZE_OCRAM 0x020000
```

kemonats (commented):
This PR is approved but I have some suggestions.
I also have a proposition to improve the `hal_memoryGetNextEntry()` function ( reduces code size about 500 bytes), but that probably needs to be done as separate PR.

Makefile
gerard5: I have verified Your suggestion and it works just fine, thanks. I appreciate Your support in this topic. However, for now, I prefer the PR approach and here is why. There is no need to extend the `ifeq`, `LD_FILES` list with the complexity of the future hierarchy defined in the `*.ldt`, by inclusion of `*.lds` files, this should work like a simple template system, take top (target) file and parse the hierarchy (using C preprocessor) of the included sub-scripts, which may differ per target. E.g. :
* `imxrt`'s can soon be extended with `FLEXRAM_CONFIG` bitmap configuration scripts based on `ocram/dtcm/itcm` maps (for the `imxrt10xx` family will be different than for `imxrt117x`).
* some common parts of `plo-ia32.lds` and `plo-arm.lds` will be shared in separate sub-scripts (I believe there will be also `plo-riscv64.lds` in near future).
* placement of the device tree

The suggested change assumes that every time someone add a sub-script, he/she would have to modify the `LD_FILES` list with these newly added sub-scripts. There is effectively one `ld`, which is the processed output, one input `ldt` (top of target), and more `lds` (sub target) are mixed `linker`, `C`, and `assembler` partials.

I would accept the addition of (from the suggestion):
```Makefile
LD_FILES := ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt
```
but, someone later may have mistakenly extended this list. The preferred way is to use `#include` to import sub scripts in the top target or sub target files.

Although, the part of `Makefile` responsible for `ld` file generation may be moved to `ld/` subdirectory, but not in this PR.

About the optimizations of the `plo` size, `hal_memoryGetNextEntry()` may be optimized (I agree), but other parts too of course not in this PR. I wanted the PR to be as simple as possible to easily switch to the `linker` scripts because of upcoming FlexSPI setup on `imxrt`'s require relocation of some parts of code to the `itcm` RAM (for the `.noxip` or `.ramfunc` section), which is the priority now.

ld/armv7a7-imx6ull.ldt
gerard5: That's actually a very good sign, and a pro for the addition of linker scripts, previously it was not detected. Actually the size of OCRAM available to `plo` is 68 KB. This was discussed in the https://github.com/phoenix-rtos/plo/pull/167#discussion_r829001059 above (which is now marked as resolved).

I need to investigate this, previously it was (0x20000) 128KB (wrong assumption due to 68KB available), now maybe we just don't fit in this area and this is now being detected by the linker after adding the linker scripts.

gerard5:
Regarding [thread/167](https://github.com/phoenix-rtos/plo/pull/167#discussion_r830518822), I pushed a patch for `imx6ull`, this change needs to be verified by the team on the hardware before merge, please confirm. @HBuczynski

HBuczynski:
> Regarding [thread/167](https://github.com/phoenix-rtos/plo/pull/167#discussion_r830518822), I pushed a patch for `imx6ull`, this change needs to be verified by the team on the hardware before merge, please confirm. @HBuczynski

HBuczynski (approved):
I have tested on imx6ull and it works fine.

PR #170, opened 2022-03-24
makes: Improve concurent build of ld scripts

gerard5:
JIRA: RTOS-173

This change improves:
* concurrent builds (local using standard `CC=*-phoenix-gcc` and networked using `CC=distcc`)
* incremental build/gen. of linker scripts
* add directory guards (as prerequisite)

This PR fixes random CI issue:
https://github.com/phoenix-rtos/phoenix-rtos-project/runs/5678692510?check_suite_focus=true

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt1176, imxrt1064, stm32l4x6, zynq7k-qemu, ia32-generic-qemu).

## Checklist:

## Special treatment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ KERNEL=1
 include ../phoenix-rtos-build/Makefile.common
 include ../phoenix-rtos-build/Makefile.$(TARGET_SUFF)
 
+LDGEN ?= $(CC)
+
 CFLAGS += $(BOARD_CONFIG) -DVERSION=\"$(VERSION)\"
 CFLAGS += -I../plo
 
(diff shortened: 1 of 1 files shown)


gerard5:
@damianloew , please confirm that this hotfix, fixes the problem reported by You.

damianloew:
> @damianloew , please confirm that this hotfix, fixes the problem reported by You.

Yes, bulding plo for `armv7a9-zynq7000` target seems to work properly with this change.

Makefile
```diff
@@ -79,12 +79,12 @@ $(PREFIX_PROG_STRIPPED)%.img: $(PREFIX_PROG_STRIPPED)%.elf
 	$(SIL)$(OBJCOPY) -O binary $< $@
 
 
-$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld:
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld: | $(OBJS)
 	@echo "GEN $(@F)"
```
kemonats, line 83: Maybe just add `mkdir`
```suggestion
	@echo "GEN $(@F)"
	mkdir -p "$(@D)"
```

gerard5: Thanks for the suggestion, but it does not change the order. It's expected that `GEN` is done before the `LD` and after `CC` objects are build. Previously `GEN` was started right after `plo` build begins (see [before](https://github.com/phoenix-rtos/plo/runs/5666535778?check_suite_focus=true#step:5:335), and with [this PR](https://github.com/phoenix-rtos/plo/runs/5680619437?check_suite_focus=true#step:5:390)). The suggestion of course fixes random `nonexistent directory` but does not fix second issue with `LD` which also manifests itself in random race condition and error. So, if `$(OBJS)` are build $`(PREFIX_O)` directory exist, then `LD` may be generated - so this PR fixes both issues.

kemonats: Something I don't understand.
Why should the `*.ld` file be generated only after all `*.o` files have been created?
I understand that the error occurs when creating the `*.elf` file ?
What kind of error?

gerard5: Yes when `elf` file is being created, the error manifests when using `distcc` to build all available targets in parallel, just adding `mkdir` does not help in case of `distcc`, it randomly stops at not existing e.g. `_build/armv7a9-zynq7000/plo/phfs/msg.o` or objects from plo/lib/, use of ordered-only prerequisites helps in that case.

kemonats: From what you wrote, it seems that the problem is with `distcc` ( with `make -j` running on a single station this error does not occur).
It creates a `*.elf` file if all the files it depends on have not yet been created.
If `distcc` behaves like this, your solution probably not guaranteed to work correctly either.
However I suggest adding `mkdir` when creating the `*.ld` file.

gerard5: Don't get me wrong. Adding `mkdir` the way you suggest is awkward for me, because it doesn't treat directory as a dependency. To be well understood, I push next change to this PR (to be squashed later), maybe it will explain more why I think so. For the moment I'm marking this discussion as resolved because it applies to the first commit, I very please discuss on review of the next commit.

Makefile
```diff
@@ -30,63 +32,69 @@ OBJS += $(addprefix $(PREFIX_O), _startc.o plo.o syspage.o)
...
 	$(SIL)$(OBJCOPY) -O binary $< $@
 
 
-$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld:
+-include $(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)*ld.d
+$(PREFIX_O)/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ld: | $(PREFIX_O)/.
 	@echo "GEN $(@F)"
-	$(SIL)$(CC) $(LDSFLAGS) -D__LINKER__ -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+	$(SIL)$(LDGEN) $(LDSFLAGS) -MP -MF $@.d -MMD -D__LINKER__ -undef -xc -E -P ld/$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt > $@
+	$(SIL)sed -i 's`.*\.o[ \t]*:`$@:`' $@.d
```
kemonats, line 86: If it should work on `Darwin` sytem.
```suggestion
	$(SIL)$(SED) -i 's`.*\.o[ \t]*:`$@:`' $@.d
```

gerard5: Thanks, added compatible `sed` line and verified to build on freebsd (shall be compatible with Darwin too).

HBuczynski (approved):

PR #197, opened 2022-09-06
Add plo image to map entries and provide section start/end symbols.

gerard5:
## Description

After the modification (stm32l4) so that the plo is loaded into RAM, the `.init`, `.text`, `.rodata` sections were not mapped (map entries), which resulted in considering this area as unallocated. While loading applications to RAM below the first RAM map-entry resulted in overwriting the `plo` image when `plo` is running from the RAM.

The `ia32` scripts have been updated in a similar way to `stm32` to be consistent.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/stm32/hal.c
+++ b/hal/armv7m/stm32/hal.c
@@ -22,6 +22,11 @@ struct {
 
 
 /* Linker symbols */
+extern char __init_start[], __init_end[];
+extern char __text_start[], __etext[];
+extern char __rodata_start[], __rodata_end[];
+extern char __init_array_start[], __init_array_end[];
+extern char __fini_array_start[], __fini_array_end[];
 extern char __ramtext_start[], __ramtext_end[];
 extern char __data_start[], __data_end[];
 extern char __bss_start[], __bss_end[];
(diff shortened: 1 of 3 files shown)


damianloew (commented):
Thanks to this change loading psh (and possibly other 'big' binaries) to ram is possible (stm32l4x6 target), I've tested it and run psh tests on this target, everything seems to work fine. Tested using loading from `flash0` device, but loading through uart seems to work correctly too (`damianloew/support_serial_loading` branch). As @gerard5  discovered, there is still problem with loading smaller binaries (~64k) like tests, because of placing them between kernel and plo. In fact currently area mapped for kernel is too small, but that's for another PR. I like this change, because it will be closer to loading tests to ram without adding second ram map :)

anglov (approved):
I would go for merging it. There are still issues with maps:
- kernel uses memory that isn't reserved so small app loaded to ram causes issues.
- flash regions isn't mapped so psh mem command for flash map isn't accurate and loading to flash will cause overwriting plo on flash.
- reusage plo stack on kernel start will work but it's dangerous

But all of this is outside of the scope of this fix.

gerard5:
That's right, for example, [`AREA_KERNEL`](https://github.com/phoenix-rtos/plo/blob/d7107a79f20884c92fbbf23e5fef27baea2de8e2/ld/armv7m4-stm32l4x6.ldt#L27) can be used for the `kernelimg` command in place of  [`<data size>`, instead of taking the wrong (bss size) value from `readelf`.](https://github.com/phoenix-rtos/phoenix-rtos-project/blob/5e60a399586c77cbf28608f7777d17f444ce0be3/_targets/build.common#L78) to reserve kernel map entry in `plo`.
Ok, I'm merging this.

PR #199, opened 2022-09-12
devices: stm32l4x6: add ram storage driver based on flash-stm32

damianloew:
## Description

 - devices: stm32l4x6: add ram storage driver based on flash-stm32
 - stm32l4x6: move FLASH defines from hal to linker template
 
## Motivation and Context

- It's needed for loading apps, previously placed in RAM memory (for example by gdb and jtag). Required for stm32l4x6-nucleo testing process.

- JIRA: PD-301

- Memory areas should be located in linker script
    
- JIRA: RTOS-245

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4x6-nucleo).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/stm32/l4/peripherals.h
+++ b/hal/armv7m/stm32/l4/peripherals.h
@@ -69,8 +69,4 @@
 #define UART4_IRQ uart4_irq
 #define UART5_IRQ uart5_irq
 
-#define FLASH_PROGRAM_1_ADDR    0x08000000
-#define FLASH_PROGRAM_2_ADDR    0x08080000
-#define FLASH_PROGRAM_BANK_SIZE (512 * 1024)
-
 #endif
--- a/devices/devs.c
+++ b/devices/devs.c
@@ -17,7 +17,7 @@
 
 #include <lib/errno.h>
 
-#define SIZE_MAJOR 4
+#define SIZE_MAJOR 5
 #define SIZE_MINOR 16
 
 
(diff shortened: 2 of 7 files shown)


HBuczynski (approved):
LGTM

hal/armv7m/stm32/l4/peripherals.h
```diff
@@ -73,4 +73,7 @@
 #define FLASH_PROGRAM_2_ADDR    0x08080000
 #define FLASH_PROGRAM_BANK_SIZE (512 * 1024)
 
+#define RAM_ADDR      0x20000000
+#define RAM_BANK_SIZE (320 * 1024)
```
gerard5, line 77: We have place for statically defined memory areas: [armv7m4-stm32l4x6.ldt](https://github.com/phoenix-rtos/plo/blob/master/ld/armv7m4-stm32l4x6.ldt).  Every statically defined memory area/region should be placed there.

Btw. not the case for this PR, but `FLASH_PROGRAM_*` should also be moved to the `*ldt` file.

damianloew: Moved in the last force-push, did You mean sth like that?

HBuczynski: I agree with @gerard5 that memory areas should be located in linker script.
I know that moving `FLASH_PROGRAM_*` define is not in the scope of this PR but you can do it in a separate commit within this PR.

damianloew: Done in second commit.

gerard5: Yes.

gerard5 (approved):
Tested on HW, works as expected.

PR #203, opened 2022-09-27
hal/armv7a/_cache.S: code improvement

kemonats:
## Description

Changes:
- using `ubfx`
- changed `mov` to `lsl` ( it is prefered in that case )
- in function `hal_dcacheInvalAll()` use of register incrementation by `1 << 30` instead of by `1`.
- since the `hal_dcacheEnable()` and the `hal_icacheEnable()` functions differ only in the mask value, a local function `cacheEnable()` was defined to which the mask value in `r1` register is passed
- cache.h : added a missing `hal_dcacheInvalAll(void)` declaration

Do these functions need to be written in assembler? 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/_cache.S
+++ b/hal/armv7a/_cache.S
@@ -18,18 +18,26 @@
 
 .arm
 
-.globl hal_dcacheEnable
-.type hal_dcacheEnable, %function
-hal_dcacheEnable:
+.type cacheEnable, %function
+cacheEnable:
 	dsb
-	mrc p15, 0, r1, c1, c0, 0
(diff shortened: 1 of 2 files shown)


agkaminski (approved):
LGTM

HBuczynski (approved):
General LGTM.
The code for cache management was derived from `phoenix-rtos-kernel` where it is written in assembler. 
It uses `CP15` registers so it has to be written in that way.

kemonats:
I was thinking of such implementation:
reading from and writing to CP15 registers using inline assembly, rest of code in C.

PR #204, opened 2022-09-30
Extend functionality of debug command and memory map attributes check.

gerard5:
## Description

Extends the functionality of the memory content analysis tool. Performs `memory range checking` against defined maps and whether the `read attribute is set`. If the memory range given in the arguments for the `dump` command is incorrect, or no read access is available, or the maps are not defined at all, the command will display masked octet value `XX` for the given range.
Additionally the selected range is printed starting form alignment down to 16 bytes + offset.

rect851-3-6-7

When used with flag `-F` memory map validation is disabled (than works like the earlier implementation).

### Additional feature

An additional option is to read disks `ia32`  hard disk or floppy disk, on `imxrt-flexspi` via IP Bus it reads NOR memory, utilizing `phfs` subsystem.

Example on ia32: `dump -r hd0 0 100` prints 100 bytes from the beginning of hard drive.

2022-10-24-174126_909x253_scrot

Example on imxrt117x: `dump -F -r flash0 0x11000 0x20` reads 32 bytes of ELF header located somewhere on flash0 reading through IP Bus, not an AHB.

2022-10-24-174339_660x96_scrot

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -5,8 +5,8 @@
  *
  * Dump memory
  *
- * Copyright 2021 Phoenix Systems
- * Author: Hubert Buczynski
+ * Copyright 2021-2022 Phoenix Systems
+ * Author: Hubert Buczynski, Gerard Swiderski
  *
  * This file is part of Phoenix-RTOS.
  *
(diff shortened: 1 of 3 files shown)


agkaminski:
These invalid accessed caused CPU exception, I guess? Otherwise it might sometimes be useful to peek at memory we're not supposed to peek at.

gerard5:
> These invalid accesses caused CPU exception, I guess? Otherwise it might sometimes be useful to peek at memory we're not supposed to peek at.

That's right, but of course I'll add `a switch to bypass` this, but it will `validate` access with defined `maps by default` like in this PR.

HBuczynski (dismissed):
Nice feature ;)
LGTM.

cmds/dump.c
```diff
@@ -17,61 +17,214 @@
...
+}
+
+
+static void print_hline(void)
+{
+	unsigned int col;
+	for (col = 0; col < 79; ++col) {
+		lib_consolePutc('-');
+	}
+	lib_printf("\n");
```
agkaminski, line 111: Just a suggestion, somewhat faster approach, we don't have to worry about stack in plo that much
```suggestion
	char buff[80];
	hal_memset(buff, '-', sizeof(buff) - 1);
	buff[sizeof(buff) - 1] = '\n';
	lib_printf(buff);
```

agkaminski (dismissed):
One small suggestion, nevertheless LGTM

cmds/dump.c
gerard5: `printf()` [underneath calls](https://github.com/phoenix-rtos/plo/blob/master/lib/printf.c#L25-L29) `lib_consolePutc()` so it's not the fastest and is stack hungry, but well ... I missed that last `lib_printf("\n")` can be replaced with `lib_consolePutc('\n')` :-) Thanks in advance for pointing it out. I'll change it like so, and other calls to `printf()` that does not use formatting I'll change to `lib_consolePuts()`, this is what compiler do for `printf`s if it detects no formating it replaces with puts/putc respectively.

HBuczynski (approved):
Tested on zedboard. Works as expected.

PR #206, opened 2022-10-03
Make possible to `erase` flash.

gerard5:
## Description

I add the functionality to erase the flash memory using the `erase` command. A working implementation is provided for all supported `i.MX RT` platforms based on the NOR memory and `FlexSPI` low-level driver. Now it is easy to extend/implement low-level `flashdrv_erase()` handlers for other targets supported by the `plo`.

plo_erase_19399034

The command as below example, will wipe the entire `flash0` device:
```
erase flash0
```

The range can be specified like:
```
erase flash0 0x201000 2088960
```
If sectors are 4096 bytes in size, `2088960 bytes (510 sectors)` will be erased as a result of the above example, starting at address `0x201000` (address is aligned down to the sector size) up to `0x3ff000` address. Length argument may be either decimal or hexadecimal value.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (`imxrt1176-nil`, `mimxrt1064-evk`).

## Checklist:

## Special treatment


phfs/phfs.h
```diff
@@ -19,6 +19,10 @@
 #include <devices/devs.h>
 #include <hal/hal.h>
 
+/* phfs_open flags */
+#define PHFS_OPEN_RAWONLY 0x80000000
```
HBuczynski, line 23: I think that there is no need to create additional define to point to the same attributes as the following enum:
https://github.com/phoenix-rtos/plo/blob/6bd8f71a3f6e27c0d28184ca4216a8d9b1255db4/phfs/phfs.c#L27

I would suggest to put the `enum` in header and use it instead of defining the new fields. On the other hand, you can use defines instead of `enum` (remove it)  in the whole `phfs` unit.

gerard5: > I think that there is no need to create additional define to point to the same attributes as the following enum:
> 
> https://github.com/phoenix-rtos/plo/blob/6bd8f71a3f6e27c0d28184ca4216a8d9b1255db4/phfs/phfs.c#L27
> 
> I would suggest to put the `enum` in header and use it instead of defining the new fields. On the other hand, you can use defines instead of `enum` (remove it) in the whole `phfs` unit.

From what I see phfs_prot_* is used internally along with phfs_* structures defined in the phfs.c file, while the public API defined in phfs.h provides phfs_open() with parameter flags. I believe the purpose of flags is to tell phfs_open() how it should behave (rather than just forward these flags to phoenixd on the remote side).

cmds/erase.c
```diff
@@ -0,0 +1,88 @@
...
+#include <devices/devs.h>
+#include <hal/hal.h>
+#include <lib/lib.h>
+
+#include <phfs/phfs.h>
+
+
+static void cmd_eraseInfo(void)
+{
+	lib_printf("erase sectors or all data from storage device using phfs interface");
```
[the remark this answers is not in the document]
gerard5: Done.

phfs/phfs.h
gerard5: Added separate commit which replaces enums with defines in the internal API.

HBuczynski (approved):

agkaminski (approved):

PR #211, opened 2022-10-18
zynq7000-flash: add erase handler implementation

HBuczynski:
## Description

Add erase handler implementation.

## Motivation and Context

`jffs2` partition needs to be erased before writing a new filesystem. 

RTOS-273

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (zynq7000-zedboard).

## Checklist:

## Special treatment

### DIFF
--- a/devices/flash-zynq7000/flashdrv.c
+++ b/devices/flash-zynq7000/flashdrv.c
@@ -187,22 +187,48 @@ static int flashdrv_welSet(unsigned int cmdID)
 }
 
 
-static int flashdrv_sectorErase(addr_t offs)
+static int flashdrv_chipErase(void)
 {
-	u32 id = 0;
-	ssize_t res;
+	int res;
+	time_t timeout;
+
+	const flash_cfi_t *cfi = &fdrv_common.info.cfi;
+	const flash_cmd_t *cmd = &fdrv_common.info.cmds[flash_cmd_be];
+
+	res = flashdrv_welSet(flash_cmd_wren);
+	if (res < 0) {
+		return res;
+	}
+
+	hal_memset(fdrv_common.cmdTx, 0, cmd->size);
+	fdrv_common.cmdTx[0] = cmd->opCode;
+
+	qspi_start();
+	res = (int)qspi_polledTransfer(fdrv_common.cmdTx, NULL, cmd->size, TIMEOUT_CMD_MS);
+	qspi_stop();
+
+	if (res < 0) {
+		return res;
+	}
+
+	timeout = CFI_TIMEOUT_MAX_ERASE(cfi->timeoutTypical.chipErase, cfi->timeoutMax.chipErase) + TIMEOUT_CMD_MS;
+
+	return flashdrv_wipCheck(timeout);
+}
+
+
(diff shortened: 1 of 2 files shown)


devices/flash-zynq7000/flashdrv.c
```diff
@@ -420,6 +447,72 @@ static ssize_t flashdrv_write(unsigned int minor, addr_t offs, const void *buff,
...
+
+	res = flashdrv_regionFind(addr, &id);
+	if (res < 0) {
+		return res;
+	}
+
+	sectSz = CFI_SIZE_SECTION(cfi->regs[id].size);
+
+	addr_mask = sectSz - 1;
+	end = (addr + len + addr_mask) & ~addr_mask;
```
agkaminski, line 470: There's no `len` sanitization - it case of `len` greater then size of the memory I guess it'll work, sectors would be erased, but we'll exit with an error

HBuczynski: Done

agkaminski (approved):

devices/flash-zynq7000/flashdrv.c
```diff
@@ -313,6 +339,7 @@ static ssize_t flashdrv_read(unsigned int minor, addr_t offs, void *buff, size_t
 	res = qspi_polledTransfer(NULL, (u8 *)buff + dataSz, len - dataSz, timeout);
 	qspi_stop();
 
+
```
gerard5, line 342: nitpick: unnecessary blank line.

devices/flash-zynq7000/flashdrv.c
```diff
@@ -187,22 +187,48 @@ static int flashdrv_welSet(unsigned int cmdID)
...
+	timeout = CFI_TIMEOUT_MAX_ERASE(cfi->timeoutTypical.chipErase, cfi->timeoutMax.chipErase) + TIMEOUT_CMD_MS;
+
+	return flashdrv_wipCheck(timeout);
+}
 
-	switch (CFI_SIZE_SECTION(cfi->regs[id].size)) {
+
+static int flashdrv_sectorErase(addr_t offs, size_t sectSz)
+{
+	ssize_t res;
```
gerard5, line 222: The result of the `flashdrv_sectorErase` function is unlikely to be of type of `ssize_t`, here only `qspi_polledTransfer()` result is of `ssize_t` type, the other returns `int`, the same for `flashdrv_chipErase(void)`.
```suggestion
	int res;
```
if so, then
```diff
-  res = qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmd->size, TIMEOUT_CMD_MS);
+  res = (int)qspi_polledTransfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmd->size, TIMEOUT_CMD_MS);TIMEOUT_CMD_MS);
```
because only negative values or zero that fall within `int` range are of interest to us, and number of erased sectors or erased size is not returned by the function.

HBuczynski: Done

gerard5 (approved):

PR #217, opened 2022-11-03
cmds: handle lf as newline character

damianloew:
## Motivation and Context

 - JIRA: PD-308

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (list targets here).

## Checklist:

## Special treatment

### DIFF
--- a/cmds/cmd.c
+++ b/cmds/cmd.c
@@ -157,6 +157,7 @@ void cmd_prompt(void)
 	char c = 0, sc = 0;
 	int pos = 0;
 	int i, k, chgfl = 0, ncl;
+	unsigned char newline = 0;
 
 	cmd_common.ll = 0;
 	cmd_common.cl = 0;
@@ -196,7 +197,17 @@ void cmd_prompt(void)
 
 		/* Regular characters */
 		if (c) {
-			if (c == '\r') {
+			if ((c == '\r') || (c == '\n')) {
+				/* handle crlf line ending */
+				if (c == '\r') {
+					newline = 1;
+				}
+				/* lf after cr - skip */
+				else if ((c == '\n') && (newline != 0)) {
+					newline = 0;
+					continue;
+				}
+
 				if (pos) {
 					cmd_parse(cmd_common.lines[cmd_common.ll]);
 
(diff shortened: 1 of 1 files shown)


agkaminski:
This might turn out problematic - some terminals are sending only CR, some only LF, but there are some (mostly Windows based) that are sending CRLF. So this solution will cause empty new line to be interpreted on every enter on such terminals.

What we can do: e.g. on CR save a flag, that CR has been read and if next char is LF ignore it.

damianloew:
> This might turn out problematic - some terminals are sending only CR, some only LF, but there are some (mostly Windows based) that are sending CRLF. So this solution will cause empty new line to be interpreted on every enter on such terminals.
> 
> What we can do: e.g. on CR save a flag, that CR has been read and if next char is LF ignore it.

Please check the solution with the `newline` flag and let me know whether it's ok

cmds/cmd.c
```diff
@@ -196,7 +197,15 @@ void cmd_prompt(void)
 
 		/* Regular characters */
 		if (c) {
-			if (c == '\r') {
+			if ((c == '\r') || (c == '\n')) {
+				if (newline) {
```
agkaminski, line 201:
```suggestion
				if (newline != 0) {
```

agkaminski (commented):
We should ignore only LR if CR was previously parsed - right now e.g. double enter won't work (we expect two empty lines, we'll get one) ;)

cmds/cmd.c
damianloew: You are right, sorry for that. The last force-push handles this case.

cmds/cmd.c
```diff
@@ -196,7 +197,17 @@ void cmd_prompt(void)
 
 		/* Regular characters */
 		if (c) {
-			if (c == '\r') {
+			if ((c == '\r') || (c == '\n')) {
+				/* handle crlf line ending */
+				if ((c == '\r') && (newline == 0)) {
```
agkaminski, line 202: `newline == 0` is redundant - we always end up with `newline` = 1 anyway
```suggestion
				if (c == '\r') {
```

agkaminski (approved):
One nitpick, otherwise LGTM

cmds/cmd.c
damianloew: Yeah, maybe that's better, changed.

agkaminski (approved):

PR #226, opened 2023-01-12
IA32 port fixes

Maxez:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/_init.S
+++ b/hal/ia32/_init.S
@@ -23,13 +23,12 @@
 .section .init
 
 
-/* Loader high level part entry point */
+/* External variables */
+.extern _plo_size
+.extern _stack
+.extern _stacksz
 .extern _startc
 .type _startc, @function
-
-
-/* Loader size */
-.extern _plo_size
 
 
 /* Loader entry point */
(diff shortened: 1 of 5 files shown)


hal/ia32/_init.S
```diff
@@ -23,15 +23,14 @@
 .section .init
 
 
-/* Loader high level part entry point */
+/* External variables */
+.extern _plo_size
```
gerard5, line 27: Add:
```suggestion
.extern _plo_size
.extern _plo_seg16
```

hal/ia32/_init.S
```diff
@@ -41,115 +40,86 @@ _start:
...
-	movl %eax, _plo_geo_secs
+	/* Set start CHS address */
+	xorw %bp, %bp
+	xorb %dh, %dh
+	movw $0x1, %ax
 
 _plo_load:
 	/* Load high level part into memory */
-	movl $_plo_size-1, %ecx
+	movw $((ADDR_PLO + 0x200) >> 4), %bx
```
gerard5, line 94: Replace with:
```suggestion
	movw $(_plo_seg16), %bx
```

gerard5 (commented):
@Maxez I like the changes, `.init` (boot sector, stage0) is much lighter now. This `$((ADDR_PLO + 0x200) >> 4)` still looks weird-strange to me, and we can take the 16bit segment from the linker script, we agreed to resign from using macrodefs as much as possible in the asm, so this offset calculation can be replaced with the linker symbol which delivers the computed 16-bit segment at which the plo `.text` (0x7e0) is loaded, and it also gives us a better way to automatically calculate the segment based on the location of `.text` we know it in the linker script, so segment can be easliy computed without constant offsets.

See my comment's suggestions (add/replace), it works pretty well after the following patch is applied. I've decided to post patch because I can't comment no existing changes to the linker script (lds) in the PR until now.

```diff
diff --git a/hal/ia32/_init.S b/hal/ia32/_init.S
index 9a35cc3..4f77165 100644
--- a/hal/ia32/_init.S
+++ b/hal/ia32/_init.S
@@ -25,6 +25,7 @@
 
 /* External variables */
 .extern _plo_size
+.extern _plo_seg16
 .extern _stack
 .extern _stacksz
 .extern _startc
@@ -66,7 +67,7 @@ _plo_start:
 
 _plo_load:
 `       /* Load high level part into memory */
-`       movw $((ADDR_PLO + 0x200) >> 4), %bx
+`       movw $(_plo_seg16), %bx
 `       movw %bx, %es
 `       xorw %bx, %bx
 `       xorb %dh, %dh
diff --git a/ld/common/plo-ia32.lds b/ld/common/plo-ia32.lds
index 992e4d8..85223fb 100644
--- a/ld/common/plo-ia32.lds
+++ b/ld/common/plo-ia32.lds
@@ -133,5 +133,8 @@ SECTIONS
 `       _plo_size = LOADADDR(.data) + SIZEOF(.data) - ORIGIN(PLO_IMAGE);
 `       _plo_load_addr = ORIGIN(PLO_IMAGE);
 
+`       /* 16bit real-mode segments */
+`       _plo_seg16 = LOADADDR(.text) >> 4;
+
 `       /DISCARD/ : { *(.note.GNU-stack) *(.gnu_debuglink) *(.gnu.lto_*) }
 }
diff --git a/ld/ia32-generic.ldt b/ld/ia32-generic.ldt
index 6e24394..c7750c1 100644
--- a/ld/ia32-generic.ldt
+++ b/ld/ia32-generic.ldt
@@ -33,8 +33,6 @@
 #define ADDR_PDIR   0x3000
 #define ADDR_PTABLE 0x4000
 
-#define ADDR_PLO 0x7c00
-
 /* Disk caches (below upper memory starting at 0x80000) */
 #define ADDR_RCACHE 0x78000
 #define SIZE_RCACHE 0x4000
```

agkaminski (commented):
I see only 2 small misra related nitpicks

agkaminski (approved):
LGTM

PR #237, opened 2023-03-25
hal/ia32: Fix kernel not starting on some hardware

astalke:
## Description

- Fix **plo** being stuck in an infinite loop, trying to enable A20 line, if A20 is enabled and there is no 8042 keyboard controller.
- Fix **plo** being stuck in an infinite loop, trying to enable A20 line, if A20 is disabled and there is no 8042 keyboard controller.
- Add an error message, when **plo** fails to enable A20 line.

DONE: RTOS-305

## Motivation and Context

**plo** didn't start the kernel on some hardware

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-pc

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/hal.c
+++ b/hal/ia32/hal.c
@@ -28,7 +28,7 @@ extern void hal_exceptionsInit(void);
 extern void hal_interruptsInit(void);
 
 /* Memory */
-extern void hal_memoryInit(void);
+extern int hal_memoryInit(void);
 
 /* Timer */
 extern void hal_timerInit(void);
@@ -58,6 +58,10 @@ void hal_init(void)
 	hal_consoleInit();
 	hal_exceptionsInit();
 	hal_interruptsInit();
-	hal_memoryInit();
+	if (hal_memoryInit() < 0) {
+		for (;;) {
+			hal_cpuHalt();
+		}
+	}
 	hal_timerInit();
 }
--- a/hal/ia32/memory.c
+++ b/hal/ia32/memory.c
@@ -195,28 +195,65 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
 
 
 /* Flushes 8042 keyboard controller buffers */
-static void memory_empty8042(void)
+static int memory_empty8042(void)
 {
 	unsigned char status;
-
-	do {
+	unsigned short int timeout;
+	for (timeout = 0xffff; timeout != 0; --timeout) {
 		/* Discard input data */
-		if ((status = hal_inb((void *)0x64)) & 0x01) {
+		status = hal_inb((void *)0x64);
+		if ((status & 0x01) != 0) {
 			hal_inb((void *)0x60);
 			continue;
 		}
-	} while (status & 0x02);
+		if ((status & 0x02) == 0) {
+			return 0;
+		}
+	}
+	return -1;
+}
+
+
+static int checkA20Line(void)
+{
+	int result = 0;
+	/* Set base to some address in the first 1MiB that can be overwritten */
+	/* Address of the magic number at the end of MBR (0xAA55) */
+	volatile u16 *const base = (void *)((ADDR_PLO) + 512 - 2);
+	/* Some values that should be equal, if A20 is off */
+	u16 v1 = *(base + (0x100000 / sizeof(*base)));
+	u16 v2 = *base;
+	if (v1 == v2) {
+		/* They are equal: try to overwrite one of them */
+		*base = v2 + 1;
+		/* Did value that shouldn't change, changed? */
+		if ((*(base + (0x100000 / sizeof(*base)))) != v1) {
+			/* A20 is off */
+			result = -1;
+		}
+		*base = v2;
+	}
+	return result;
 }
 
 
 /* Enables A20 line mask */
-static void memory_enableA20(void)
+static int memory_enableA20(void)
 {
-	memory_empty8042();
-	hal_outb((void *)0x64, 0xd1);
-	memory_empty8042();
-	hal_outb((void *)0x60, 0xdf);
-	memory_empty8042();
+	if (checkA20Line() < 0) {
+		if (memory_empty8042() < 0) {
+			return -1;
+		}
+		hal_outb((void *)0x64, 0xd1);
+		if (memory_empty8042() < 0) {
+			return -1;
+		}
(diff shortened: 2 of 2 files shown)


hal/ia32/memory.c
```diff
@@ -195,28 +195,69 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
+	unsigned short int v2 = *(unsigned short int *)base;
+	if (v1 == v2) {
+		/* They are equal: try to overwrite one of them */
+		*(int *)base = v2 + 1;
+		/* Did value that shouldn't change, changed? */
+		if ((*(int *)(base + 0x100000)) != v1) {
+			/* A20 is off */
+			result = -1;
+		}
+		*(int *)base = v2;
```
agkaminski, line 233: If it was safe to access this memory, we shouldn't need to restore it's value
```suggestion
```

hal/ia32/memory.c
```diff
@@ -228,7 +269,10 @@ int hal_memoryAddMap(addr_t start, addr_t end, u32 attr, u32 mapId)
 
 int hal_memoryInit(void)
 {
-	memory_enableA20();
+	if (memory_enableA20() < 0) {
+		hal_consolePrint("Failed to enable A20 line\n");
```
agkaminski, line 273:
```suggestion
		hal_consolePrint("hal: Failed to enable A20 line\n");
```

hal/ia32/memory.c
```diff
@@ -195,28 +195,69 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
-static void memory_enableA20(void)
+static int memory_enableA20(void)
 {
-	memory_empty8042();
-	hal_outb((void *)0x64, 0xd1);
-	memory_empty8042();
-	hal_outb((void *)0x60, 0xdf);
-	memory_empty8042();
+	if (checkA20Line() < 0) {
+		__asm__ volatile("cli");
```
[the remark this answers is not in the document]
agkaminski: Anyway, why disable interrupts here? I think interrupts aren't even enabled at this point, so "re-enabling" is actually a bug.

Interrupts are enabled by accident by `timer_init()` - there's no check in `hal_interruptsSet()` and it re-enables interrupts always (even if were disabled).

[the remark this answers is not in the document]
astalke: I've just checked. Interrupts were disabled there, so it is indeed a bug.

hal/ia32/memory.c
```diff
@@ -195,28 +195,69 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
+{
+	int result = 0;
+	/* Set base to some address in the first 1MiB that can be overwritten */
+	void *const base = (void *)0x7DFE;
+	/* Some values that should be equal, if A20 is off */
+	unsigned short int v1 = *(unsigned short int *)(base + 0x100000);
+	unsigned short int v2 = *(unsigned short int *)base;
+	if (v1 == v2) {
+		/* They are equal: try to overwrite one of them */
+		*(int *)base = v2 + 1;
```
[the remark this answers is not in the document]
astalke: I forgot to change it, while I was changing `int` to `short int`. I've also just realized that `u16` is defined there, so I'll use that instead of `unsigned short int`.

hal/ia32/memory.c
```diff
@@ -195,28 +195,69 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
+	}
+	return -1;
+}
+
+
+static int checkA20Line(void)
+{
+	int result = 0;
+	/* Set base to some address in the first 1MiB that can be overwritten */
+	void *const base = (void *)0x7DFE;
```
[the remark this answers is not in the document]
astalke: It is the magic number at the of of loaded MBR (0xAA55) and as far as I am aware, it is not overwritten at that moment of boot, but I may be mistaken.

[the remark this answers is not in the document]
agkaminski: It is somewhat common knowledge as it turns out, but IMHO comment what this address is there would be nice

hal/ia32/memory.c
```diff
@@ -195,28 +195,64 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
+	}
+	return -1;
+}
+
+
+static int checkA20Line(void)
+{
+	int result = 0;
+	/* Set base to some address in the first 1MiB that can be overwritten */
+	volatile u16 *const base = (void *)0x7dfe; /* Address of the magic number at the end of MBR (0xAA55) */
```
gerard5, line 221: `ADDR_PLO` is `0x7c00` where MBR (one sector 512 bytes) is loaded.
```suggestion
	/* Address of the magic number at the end of MBR (0xAA55) */
	volatile u16 *const base = (void *)((ADDR_PLO) + 512 - 2);
```

hal/ia32/hal.c
```diff
@@ -58,6 +58,8 @@ void hal_init(void)
 	hal_consoleInit();
 	hal_exceptionsInit();
 	hal_interruptsInit();
-	hal_memoryInit();
+	if (hal_memoryInit() < 0) {
+		for (;;) { }
```
gerard5, line 62: Let's not heat the processor to the max, let it rest.
```suggestion
		for (;;) { 
			hal_cpuHalt();
		}
```

hal/ia32/memory.c
```diff
@@ -195,28 +195,64 @@ int hal_memoryGetNextEntry(addr_t start, addr_t end, mapent_t *entry)
...
 
 /* Flushes 8042 keyboard controller buffers */
-static void memory_empty8042(void)
+static int memory_empty8042(void)
 {
 	unsigned char status;
-
-	do {
+	unsigned short int timeout;
+	for (timeout = 0xffff; timeout > 0; --timeout) {
```
gerard5, line 202: `timeout` is never `<0`, because it's `unsigned`
```suggestion
	for (timeout = 0xffff; timeout != 0; --timeout) {
```

agkaminski: It's ok as is, timeout > 0 means the same thing as timeout != 0 in thus case

agkaminski (approved):

PR #240, opened 2023-04-19
imxrt117x: Use default uart setup for imxrt1176-evk

gerard5:
`UART1` is the default console port for imxrt1176-evk to which FreeLink virtual comm. port (USB ACM-CDC interface) is connected. This setting may be overridden in board config by `UART_CONSOLE` define or if plo console should use dedicated uart in such case `UART_CONSOLE_PLO` in board config may be used.

JIRA: NIL-422

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/imxrt/117x/peripherals.h
+++ b/hal/armv7m/imxrt/117x/peripherals.h
@@ -73,13 +73,12 @@
 #define UART12 0
 #endif
 
-#ifdef UART_CONSOLE_PLO
-#undef UART_CONSOLE
-#define UART_CONSOLE UART_CONSOLE_PLO
-#endif
-
 #ifndef UART_CONSOLE
 #define UART_CONSOLE 1
+#endif
+
+#ifndef UART_CONSOLE_PLO
+#define UART_CONSOLE_PLO UART_CONSOLE
 #endif
 
 #define UART_CLK      24000000
--- a/hal/armv7m/imxrt/117x/peripherals.h
+++ b/hal/armv7m/imxrt/117x/peripherals.h
@@ -26,7 +26,7 @@
 
 /* UART */
 #ifndef UART1
-#define UART1 0
+#define UART1 1
 #endif
 
 #ifndef UART2
@@ -73,8 +73,13 @@
 #define UART12 1
 #endif
 
+#ifdef UART_CONSOLE_PLO
+#undef UART_CONSOLE
+#define UART_CONSOLE UART_CONSOLE_PLO
+#endif
+
 #ifndef UART_CONSOLE
-#define UART_CONSOLE 11
+#define UART_CONSOLE 1
 #endif
 
 #define UART_CLK      24000000
--- a/hal/armv7m/imxrt/117x/peripherals.h
+++ b/hal/armv7m/imxrt/117x/peripherals.h
@@ -66,11 +66,11 @@
 #endif
 
 #ifndef UART11
-#define UART11 1
+#define UART11 0
 #endif
 
 #ifndef UART12
-#define UART12 1
+#define UART12 0
 #endif
 
 #ifdef UART_CONSOLE_PLO
--- a/hal/armv7m/imxrt/10xx/console.c
+++ b/hal/armv7m/imxrt/10xx/console.c
@@ -13,6 +13,8 @@
  * %LICENSE%
  */
 
+#include <board_config.h>
+
 #include <hal/hal.h>
 
 #define CONCAT(a, b)         a##b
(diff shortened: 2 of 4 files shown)


hal/armv7m/imxrt/117x/peripherals.h
```diff
@@ -66,15 +66,20 @@
...
 #endif
 
 #ifndef UART12
-#define UART12 1
+#define UART12 0
+#endif
+
+#ifdef UART_CONSOLE_PLO
+#undef UART_CONSOLE
+#define UART_CONSOLE UART_CONSOLE_PLO
```
agkaminski, line 78: Perhaps plo should use `UART_CONSOLE_PLO` instead and here do
```
#ifndef UART_CONSOLE_PLO
#define UART_CONSOLE_PLO UART_CONSOLE
#endif
```

gerard5: I see `undef` is no-go here, but it's the same as in imxrt1064:
https://github.com/phoenix-rtos/plo/blob/dc68030b9ee8cc21eab861b6b274368fa31dddf1/hal/armv7m/imxrt/10xx/peripherals.h#L91-L100

I agree with your suggestion. I can change this for imxrt 1064 as well but it requires replacing all occurrences of `UART_CONSOLE` with `UART_CONSOLE_PLO` in [`console.c`](https://github.com/phoenix-rtos/plo/blob/master/hal/armv7m/imxrt/10xx/console.c) implementation - for me it is ok - I may do it if you agree ?

agkaminski: I don't mean it as no-go, I just think different approach seems better.

This approach with `UART_CONSOLE_PLO` is used on zynq target already, so I see nothing of controversy here, if you're willing to make the changes I'd be grateful ;)

agkaminski (approved):
LGTM

agkaminski:
One small thought though - we don't really need `UART_CONSOLE` define in plo now, do we?

gerard5:
> One small thought though - we don't really need `UART_CONSOLE` define in plo now, do we?

What if `UART_CONSOLE_PLO` is not defined? `UART_CONSOLE` is used as fall back, and what if `UART_CONSOLE` is not defined ?

That's why I opt to stay with:

```C
/*
 * needed as system-wide fallback to UART_CONSOLE
 * (if neither UART_CONSOLE nor UART_CONSOLE is defined) in UART_CONSOLE_PLO
 */
#ifndef UART_CONSOLE
#define UART_CONSOLE 1
#endif

/* as in your suggestion: */
#ifndef UART_CONSOLE_PLO
#define UART_CONSOLE_PLO UART_CONSOLE
#endif
```

agkaminski:
> > One small thought though - we don't really need `UART_CONSOLE` define in plo now, do we?
> 
> What if `UART_CONSOLE_PLO` is not defined? `UART_CONSOLE` is used as fall back, and what if `UART_CONSOLE` is not defined ?
> 
> That's why I opt to stay with:
> 
> ```c
> /*
>  * needed as system-wide fallback to UART_CONSOLE
>  * (if neither UART_CONSOLE nor UART_CONSOLE is defined) in UART_CONSOLE_PLO
>  */
> #ifndef UART_CONSOLE
> #define UART_CONSOLE 1
> #endif
> 
> /* as in your suggestion: */
> #ifndef UART_CONSOLE_PLO
> #define UART_CONSOLE_PLO UART_CONSOLE
> #endif
> ```

What I suggest instead is:
```
#ifndef UART_CONSOLE_PLO
#define UART_CONSOLE_PLO 1
#endif
```

agkaminski:
Hmmm, ok, your version is superior - it uses `UART_CONSOLE` if provided

gerard5:
> > > One small thought though - we don't really need `UART_CONSOLE` define in plo now, do we?
> > 
> > 
> > What if `UART_CONSOLE_PLO` is not defined? `UART_CONSOLE` is used as fall back, and what if `UART_CONSOLE` is not defined ?
> > That's why I opt to stay with:
> > ```c
> > /*
> >  * needed as system-wide fallback to UART_CONSOLE
> >  * (if neither UART_CONSOLE nor UART_CONSOLE is defined) in UART_CONSOLE_PLO
> >  */
> > #ifndef UART_CONSOLE
> > #define UART_CONSOLE 1
> > #endif
> > 
> > /* as in your suggestion: */
> > #ifndef UART_CONSOLE_PLO
> > #define UART_CONSOLE_PLO UART_CONSOLE
> > #endif
> > ```
> 
> What I suggest instead is:
> 
> ```
> #ifndef UART_CONSOLE_PLO
> #define UART_CONSOLE_PLO 1
> #endif
> ```

Now that would default to use uart `1` _always_ if` UART_CONSOLE_PLO` is not defined, what if single project-wide `UART_CONSOLE` is used to indicate both `system` and `plo` console (on NIL we use `UART_CONSOLE=11` for both plo and sysem)

PR #252, opened 2023-05-12
GR716: Implement plo copy to SRAM

lukileczo:
## Motivation and Context

JIRA: RTOS-450
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `sparcv8leon3-gr716-mini`

## Checklist:

## Special treatment

### DIFF
--- a/hal/sparcv8leon3/_init.S
+++ b/hal/sparcv8leon3/_init.S
@@ -18,11 +18,118 @@
 #include "config.h"
 #include "../cpu.h"
 
+.extern _plo_load_addr
+.extern _plo_size
+
 .section ".init", "ax"
 .align 4
 .global _init
 .type _init, #function
 _init:
+	/* get current PC */
+	call . + 8
+	nop
+	mov %o7, %g5
+	/* check where we're loaded */
+	set _plo_load_addr, %g2
+	cmp %g5, %g2
+	bge stage1
+
+	/* Enable CGU for SRAM */
+	set 0x80006000, %g1 /* %g1 = cgu_unlock */
+	add %g1, 0x8, %g3   /* %g3 = cgu_core_reset */
+	add %g1, 0x4, %g4   /* %g4 = cgu_clk_en */
+	ld [%g1], %g2
+	or %g2, 0x10, %g2
+	st %g2, [%g1]
+
+	ld [%g3], %g2
+	or %g2, 0x10, %g2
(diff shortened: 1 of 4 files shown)


hal/sparcv8leon3/_init.S
```diff
@@ -18,11 +18,118 @@
...
+.extern _plo_load_addr
+.extern _plo_size
+
 .section ".init", "ax"
 .align 4
 .global _init
 .type _init, #function
 _init:
+	/* get current PC */
+	call . + 8
```
agkaminski, line 30: Won't this cause us to jump to SRAM (regardless if it was initialized and if plo has been copied there)? `.` is provided by the compiler and it thinks we're executing from SRAM.

hal/sparcv8leon3/_traps.S
```diff
@@ -43,9 +43,9 @@
 
 #define RESET \
 	mov	%g0, %g4; \
-	sethi %hi(_init), %g4; \
-	jmp %g4 + %lo(_init); \
-	nop;
+	nop; \
+	call _init; \
```
agkaminski, line 47: Why replace with call?

Side note - this is somewhat riscy strategy, same HW can be in an unexpected state after this kind of warm reset. Is there's no way to generate MCU reset in software?

hal/sparcv8leon3/_traps.S
```diff
@@ -43,9 +43,9 @@
...
+	nop; \
```
agkaminski, line 48:
```suggestion
	nop;
```

hal/sparcv8leon3/gr716/gr716.c
```diff
@@ -139,7 +139,7 @@ static void _gr716_pllSetDefault(void)
 
 
 /* Initialize external SRAM */
-static void _gr716_sram_init(void)
+void _gr716_sram_init(void)
```
agkaminski, line 142: I guess it became a dead code now, init is done in asm

hal/sparcv8leon3/_traps.S
lukileczo: call is pc relative, so this allows for correct jump even when we're loaded to incorrect address

agkaminski: It is indeed, you're right

agkaminski (approved):
LGTM

PR #291, opened 2023-09-19
gr716: fix filter-out linker flags

lukileczo:
JIRA: RTOS-613

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `sparcv8leon3-gr716-mini`

## Checklist:

## Special treatment

### DIFF
--- a/hal/sparcv8leon3/gr716/Makefile
+++ b/hal/sparcv8leon3/gr716/Makefile
@@ -6,9 +6,11 @@
 # %LICENSE%
 #
 
+comma := ,
+
 LDFLAGS:=$(filter-out -Tbss% , $(LDFLAGS))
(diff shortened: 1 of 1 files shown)


nalajcie (approved):
Seems ok, haven't tested it.

You may want to use `LDFLAGS_PREFIX` if this would be a more generic file instead (might work also if there would be some targets using `ld` instead of `gcc`) but current approach is good also.

IMHO this should be redesigned (but not as this PR):
 - kernel-specific defines should not be appended to CFLAGS in `phoenix-rtos-build` but in `kernel`)
 - the define should not be named `KERNEL` but eg. `BARE_METAL` or `FREESTANDING` or something like that (to indicate we're compiling applications not intended to be run under phoenix-rtos-kernel control)
 - it would be better to be able to specify float ABI explicitly (and use the value in target/*.mk), not filter-out the result

PR #295, opened 2023-10-11
jffs2: log progress

agkaminski:
DONE: RTOS-627

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/cmds/jffs2.c
+++ b/cmds/jffs2.c
@@ -145,7 +145,11 @@ static int cmd_jffs2(int argc, char *argv[])
 
 	cleanmarker.hdr_crc = lib_crc32((const u8 *)&cleanmarker, sizeof(struct jffs2_cleanmarker_node) - 4u, 0);
 
+	log_info("\n");
 	for (i = 0; i < numBlocks; i++) {
+		/* FIXME should be: log_info("\rjffs2: block %lu/%lu", i, numBlocks); */
+		lib_printf("\rjffs2: block %lu/%lu", i, numBlocks);
+
 		if (erase != 0) {
 			err = devs_erase(major, minor, (startBlock + i) * blockSize, blockSize, 0);
 			if (err < 0) {


cmds/jffs2.c
```diff
@@ -145,7 +145,10 @@ static int cmd_jffs2(int argc, char *argv[])
 
 	cleanmarker.hdr_crc = lib_crc32((const u8 *)&cleanmarker, sizeof(struct jffs2_cleanmarker_node) - 4u, 0);
 
+	log_info("\n");
 	for (i = 0; i < numBlocks; i++) {
+		log_info("\rjffs2: block %lu/%lu", i, numBlocks);
```
damianloew, line 150: `log_info` seems to be more relevant, but I'd opt for using lib_printf as in `erase` to show those logs within CI (we don't have echo enabled by default and we prefer not to see all the logs). Those are crucial to see the progress of erase process.

agkaminski: @nalajcie Your thoughts?

damianloew: Hmm, alternatively we could set echo in test runner. Please choose the approach you prefer more. If we chose log_info, it would be good to apply it in all flash drivers (now there are both lib_printf and log_info)

agkaminski: It's ok to mix printf and log, I think we want to disable only some logs when echo is off.

Changed to printf for now, please take a look

nalajcie: I don't have a preference - either approaches are ok for me. I don't use `plo` often enough to have a say in having these logs enabled by default or not :)

damianloew: > It's ok to mix printf and log, I think we want to disable only some logs when echo is off.
> 
> Changed to printf for now, please take a look

Looks good to me. I meant that the same function (flashdrv_erase) has sometimes log_info and sometimes lib_printf. The only difference is target flash memory, do you want to leave it as it is?

damianloew (approved):

PR #308, opened 2023-12-04
Implement Real Time Transfer (RTT pipe) device driver

gerard5:
## Description

`Real Time Transfer`, a.k.a. `RTT` is an interface specified by SEGGER and implemented in their tools and freely implemented in OpenOCD. It is based on basic memory reads and writes to transfer data bidirectionally between target and host. The specification is independent of the target architecture. Every target that supports so called _"background memory access"_, which means that the target memory can be accessed by the debugger while the target is running, can be used. This interface is the fastest alternative for semihosting. It's speed is determined by the debug probe speed, (this implementation was tested using 12MHz and 24MHz JTAG probe; 6MHz and 4MHz SWD).

This change implements two RTT channels:
1) `console` which is used `in parallel to UART console` (but through debug interface)
2) `auxiliary pipe` for `phfs` protocol used by `phoenixd` server

To use with OpenOCD, the setup is:
```tcltk
rtt setup 0x20057f00 256 "RTT PIPE"
rtt server start 18021 0
rtt server start 18022 1
rtt polling_interval 20
```

Which initializes two TCP servers on `18021` and `18022` ports which corresponds to two channels described above (`console` and `phoenixd`).

*NOTE: Given in example `0x2003ff00` address (for rt1176) is set using macro define `RTT_ADDR` in `peripheral.h`, and might be overridden in per project `board_config.h`, the address provides location of RTT descriptors (last 256 bytes of DTCM)
Default RTT tag `"RTT PIPE"` may be overridden using `RTT_TAG_BACKWARD` define (written backwards).

To connect phoenixd run (imxrt117x is an example target):
```
phoenixd -s _boot/armv7m7-imxrt117x-evk/ -t 127.0.0.1:18022
```

Depending on configuration of OpenOCD, any IP address can be specified and remote programming and console access is possible.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176-nil`, `imxrt1064-evk`, `stm32l4a6-nucleo`

## Checklist:

## Special treatment

### DIFF
--- a/devices/pipe-rtt/device.c
+++ b/devices/pipe-rtt/device.c
@@ -1,0 +1,97 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system loader
+ *
+ * RTT pipes: communication through debug probe (device)
+ *
+ * Copyright 2023-2024 Phoenix Systems
+ * Author: Gerard Swiderski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <devices/devs.h>
+#include <hal/hal.h>
+#include <lib/lib.h>
+
+#include "rtt.h"
+
+static ssize_t pipe_read(unsigned int minor, addr_t offs, void *buff, size_t len, time_t timeout)
+{
+	(void)offs;
+	(void)timeout;
+
+	return rtt_read(minor + 1, buff, len);
+}
+
+
+static ssize_t pipe_write(unsigned int minor, addr_t offs, const void *buff, size_t len)
+{
+	(void)offs;
+
+	return rtt_write(minor + 1, buff, len);
+}
+
+
+static int pipe_done(unsigned int minor)
+{
+	return rtt_check(minor + 1);
+}
+
+
+static int pipe_sync(unsigned int minor)
+{
+	return rtt_check(minor + 1);
+}
+
+
+static int pipe_map(unsigned int minor, addr_t addr, size_t sz, int mode, addr_t memaddr, size_t memsz, int memmode, addr_t *a)
+{
+	(void)addr;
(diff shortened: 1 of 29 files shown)


devices/pipe-rtt/device.c
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+#include <devices/devs.h>
+#include <hal/hal.h>
+#include <lib/lib.h>
+
+
+extern int rtt_check(int chan);
+extern ssize_t rtt_write(int chan, const void *buf, size_t count);
+extern ssize_t rtt_read(int chan, void *buf, size_t count);
```
agkaminski, line 23: Why not provide these prototypes in a header file?

devices/pipe-rtt/rtt.c
```diff
@@ -0,0 +1,207 @@
...
+
+	rtt->txChannel[chan].wr = wr;
+
+	hal_cpuDataMemoryBarrier();
+
+	return count - todo;
+}
+
+
+ssize_t rtt_writeBlocking(int chan, const void *buf, size_t count)
```
agkaminski, line 136:
```suggestion
static ssize_t rtt_writeBlocking(int chan, const void *buf, size_t count)
```

agkaminski (commented):
Small nitpicks, I really like `*Hook` approach :)

devices/pipe-rtt/rtt.c
gerard5: Done.

devices/pipe-rtt/device.c
gerard5: Done.

agkaminski (approved):
LGTM

PR #327, opened 2024-05-08
cmd: Add blob command

agkaminski:
This command allows to pass any binary blob to the kernel as a syspage program. Kernel might be unable to execute it of course, but it will not try to, unless asked to do it (i.e. -x flag is not present for blob in the syspage).

Blob can be placed in any memory map, it will be copied by the plo to the destination map, if source != dest.

DONE: NIL-571

## Description

Building scripts need modification for this cmd to be usable. Example modification:

```
diff --git a/_targets/build.common b/_targets/build.common
index 2ddc966..2318bb1 100644
--- a/_targets/build.common
+++ b/_targets/build.common
@@ -69,6 +69,14 @@ b_mkscript_user() {
                                        ((poffs+=(psz + SIZE_PAGE - 1) & PAGE_MASK))
                                        printf "%s\n" "$cmd";;
 
+                               blob)
+                                       name=${args[2]%%;*}
+                                       PROGS+=("$name")
+                                       psz=$(wc -c < "$name")
+                                       printf "alias%s %s 0x%x 0x%x\n" "$RELATIVE_ALIAS" "$name" "$poffs" "$psz"
+                                       ((poffs+=(psz + SIZE_PAGE - 1) & PAGE_MASK))
+                                       printf "%s\n" "$cmd";;
+
                                kernel)
                                        printf "alias%s %s 0x%x 0x%x\n" "$RELATIVE_ALIAS" "${KERNEL_FILE}" "$KERNEL_OFFS" "$ksz"
                                        printf "%s\n" "$cmd";;

```
Usage: add blob command to the user script, example:

```
diff --git a/_targets/build.project.armv7m4-stm32l4x6 b/_targets/build.project.armv7m4-stm32l4x6
index c0bf5fa..9e794be 100644
--- a/_targets/build.project.armv7m4-stm32l4x6
+++ b/_targets/build.project.armv7m4-stm32l4x6
@@ -70,6 +70,7 @@ USER_SCRIPT=(
        "kernelimg ${BOOT_DEVICE}"
        "app ${BOOT_DEVICE} -x stm32l4-multi flash0 ram"
        "app ${BOOT_DEVICE} -x psh flash0 ram"
+       "blob ${BOOT_DEVICE} blob ram"
```

blob will appear in the /syspage/ dir on the target and be usable as a read-only dummyfs file.
blob file will have execute flag set by the dummyfs, but it doesn't hurt anything, I guess

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4).

## Checklist:

## Special treatment

### DIFF
--- a/cmds/blob.c
+++ b/cmds/blob.c
@@ -1,0 +1,170 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system loader
+ *
+ * Load application
+ *
+ * Copyright 2020-2021, 2024 Phoenix Systems
+ * Author: Hubert Buczynski, Gerard Swiderski, Aleksander Kaminski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "cmd.h"
+
+#include <lib/lib.h>
+#include <hal/hal.h>
+#include <phfs/phfs.h>
+#include <syspage.h>
+
+
+static void cmd_blobInfo(void)
+{
+	lib_printf("put file in the syspage, usage: blob [<dev> <name> <map>]");
+}
+
+static int cmd_cp2ent(handler_t handler, const mapent_t *entry)
+{
+	ssize_t len;
+	u8 buff[SIZE_MSG_BUFF];
+	size_t offs, sz = entry->end - entry->start;
+
+	for (offs = 0; offs < sz; offs += len) {
+		len = phfs_read(handler, offs, buff, min(SIZE_MSG_BUFF, sz - offs));
+		if (len < 0) {
+			log_error("\nCan't read data");
+			return len;
+		}
+		hal_memcpy((void *)(entry->start + offs), buff, len);
+	}
+
+	return EOK;
+}
+
+static int cmd_blobLoad(handler_t handler, size_t size, const char *name, const char *map)
+{
+	int res;
+
+	unsigned int attr;
+	addr_t start, end, offs, addr;
+
+	syspage_prog_t *prog;
+	const mapent_t *entry;
+
+	if (phfs_aliasAddrResolve(handler, &offs) < 0) {
+		offs = 0;
+	}
+
+	if (syspage_mapAttrResolve(map, &attr) < 0 ||
+			syspage_mapRangeResolve(map, &start, &end) < 0) {
+		log_error("\n%s does not exist", map);
+		return -EINVAL;
+	}
+
+	/* Check whether map's range coincides with device's address space */
+	res = phfs_map(handler, offs, size, mAttrRead, start, end - start, attr, &addr);
+	if (res < 0) {
+		log_error("\nDevice is not mappable in %s", map);
+		return res;
+	}
+
+	if (res == dev_isMappable) {
+		entry = syspage_entryAdd(NULL, addr + offs, size, SIZE_PAGE);
+		if (entry == NULL) {
+			log_error("\nCannot allocate memory for %s", name);
+			return -ENOMEM;
+		}
+	}
+	else if (res == dev_isNotMappable) {
+		entry = syspage_entryAdd(map, (addr_t)-1, size, SIZE_PAGE);
+		if (entry == NULL) {
+			log_error("\nCannot allocate memory for %s", name);
+			return -ENOMEM;
+		}
+
+		/* Copy file to the selected entry */
+		res = cmd_cp2ent(handler, entry);
+		if (res < 0) {
+			return res;
+		}
+	}
+	else {
+		log_error("\nDevice mappable routine failed");
+		return -ENOMEM;
+	}
+
+	prog = syspage_progAdd(name, 0);
+	if (prog == NULL) {
+		log_error("\nCannot add syspage program for %s", name);
+	}
+
+	prog->imaps = NULL;
+	prog->imapSz = 0;
+	prog->dmaps = NULL;
+	prog->dmapSz = 0;
+	prog->start = entry->start;
+	prog->end = entry->end;
+
+	return EOK;
+}
+
+
+static int cmd_blob(int argc, char *argv[])
+{
+	int res;
+	const char *dev;
+	const char *name;
+	const char *map;
+
+	handler_t handler;
+	phfs_stat_t stat;
+
+	/* Parse command arguments */
+	if (argc == 1) {
+		syspage_progShow();
+		return CMD_EXIT_SUCCESS;
+	}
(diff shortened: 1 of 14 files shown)


cmds/blob.c
```diff
@@ -0,0 +1,167 @@
...
+}
+
+
+static int cmd_blob(int argc, char *argv[])
+{
+	int res;
+
+	const char *dev = argv[1];
+	const char *name = argv[2];
+	const char *map = argv[3];
```
lukileczo, line 121: Technically you access `argv` elements before checking it's size

cmds/blob.c
```diff
@@ -0,0 +1,167 @@
...
+	}
+	else if (argc != 4) {
+		log_error("\n%s: Wrong argument count", argv[0]);
+		return CMD_EXIT_FAILURE;
+	}
+
+	/* Open file */
+	res = phfs_open(dev, name, 0, &handler);
+	if (res < 0) {
+		log_error("\nCan't open %s on %s (%d)", name, argv[1], res);
```
lukileczo, line 139:
```suggestion
		log_error("\nCan't open %s on %s (%d)", name, dev, res);
```

cmds/blob.c
```diff
@@ -0,0 +1,167 @@
...
+	res = phfs_stat(handler, &stat);
+	if (res < 0) {
+		log_error("\nCan't get stat from %s (%d)", name, res);
+		phfs_close(handler);
+		return CMD_EXIT_FAILURE;
+	}
+
+	res = cmd_blobLoad(handler, stat.size, name, map);
+	if (res < 0) {
+		log_error("\nCan't load %s to %s via %s (%d)", name, map, argv[1], res);
```
lukileczo, line 153:
```suggestion
		log_error("\nCan't load %s to %s via %s (%d)", name, map, dev, res);
```

cmds/blob.c
```diff
@@ -0,0 +1,167 @@
...
+
+	handler_t handler;
+	phfs_stat_t stat;
+
+	/* Parse command arguments */
+	if (argc == 1) {
+		syspage_progShow();
+		return CMD_EXIT_SUCCESS;
+	}
+	else if (argc != 4) {
```
lukileczo, line 131: Redundant `else` (and MISRA will be happy)
```suggestion
	if (argc != 4) {
```

cmds/blob.c
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+	else if (res == dev_isNotMappable) {
+		entry = syspage_entryAdd(map, (addr_t)-1, size, SIZE_PAGE);
+		if (entry == NULL) {
+			log_error("\nCannot allocate memory for %s", name);
+			return -ENOMEM;
+		}
+
+		/* Copy elf file to selected entry */
```
lukileczo, line 88:
```suggestion
		/* Copy file to selected entry */
```

lukileczo (dismissed):
LGTM

lib/cbuffer.c
```diff
@@ -17,7 +17,7 @@
 #include "cbuffer.h"
 
 
-size_t lib_cbufSize(const cbuffer_t *buf)
+__attribute__((section(".noxip"))) size_t lib_cbufSize(const cbuffer_t *buf)
```
lukileczo, line 20: The `.noxip` section is defined in linker script only on arm. On other archs it's an [orphan section](https://sourceware.org/binutils/docs/ld/Orphan-Sections.html) (and it's up to linker to do sth with it). Even though it's handled correctly (I've checked readelf on sparc) I think it'd be safer to place this section explicitly on other archs in `.text` especially since we rely on `__etext` symbol.

cmds/blob.c
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	log_info("\nLoaded %s", name);
+	phfs_close(handler);
+
+	return CMD_EXIT_SUCCESS;
+}
+
+
+static const cmd_t app_cmd __attribute__((section("commands"), used)) = {
```
lukileczo, line 168:
```suggestion
static const cmd_t blob_cmd __attribute__((section("commands"), used)) = {
```

agkaminski: Thanks!

lukileczo (approved):
LGTM

Darchiv (commented):
On a target everything is fine, the blob is put into syspage and can be read as file, etc.

I'm curious, though, how this command is meant to be used. You have provided an example diff for phoenix-rtos-project/_targets/build.common to incorporate the command into the build system, but from where will the blob file be read? Should it reside in `_build/*/prog.stripped`, in `_build/*/`, some project-defined directory, or relative to `$PREFIX_PROJECT`?

This command may be very useful for some static data, media, config files, etc. which don't necessarily need to be copied to build directory, so maybe something like `blob ${BOOT_DEVICE} phoenix-rtos-project/data/config.bin ram` is a viable option? These files may need to be stripped from basedir before being put into syspage, but you get the idea.

hal/armv7m/imxrt/10xx/106x/Makefile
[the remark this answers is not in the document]
agkaminski: Missed that one, added

agkaminski:
> On a target everything is fine, the blob is put into syspage and can be read as file, etc.
> 
> I'm curious, though, how this command is meant to be used. You have provided an example diff for phoenix-rtos-project/_targets/build.common to incorporate the command into the build system, but from where will the blob file be read? Should it reside in `_build/*/prog.stripped`, in `_build/*/`, some project-defined directory, or relative to `$PREFIX_PROJECT`?
> 
> This command may be very useful for some static data, media, config files, etc. which don't necessarily need to be copied to build directory, so maybe something like `blob ${BOOT_DEVICE} phoenix-rtos-project/data/config.bin ram` is a viable option? These files may need to be stripped from basedir before being put into syspage, but you get the idea.

My example is just an easy way to test this feature with our current build scripts.
To use this feature our building scripts need to change, some per-project place for blobs needs to be created etc. As blobs most likely won't be build during project build, I think placement in _project directory would be most logical.

nalajcie:
I think we should assume the target rootfs to be the `staging` from where we're preparing the final partitions (this approach is used eg. in CI where we're passing the target rootfs tarball on NOMMU targets and load binaries one-by-one). Maaaybe we should change the implementation of building scripts to use the rootfs files directly also? 

To have project-specific blobs installed in rootfs - you can use per-project rootfs overlay - so this is already taken care of.

Darchiv (approved):

PR #330, opened 2024-05-13
imxrt-flash/imxrt117x: enable second flash memory & bug fixes

ziemleszcz:
- probe second flash memory on port B1
 - set FLEXSPI2 clock to 132 MHz
 - initial flash size fix
 - flexspi_getAddressByPort() fix
 - change FLEXSPI1 SS0 pin on port B1 to SD_B1_4 (SD_B2_05 is already used for DQS on port A1)
 - fix _imxrt_setIOmux() parameters (mode and sion were switched)
 - code style changes

JIRA: NIL-570

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt1176).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/imxrt/10xx/105x/peripherals.h
+++ b/hal/armv7m/imxrt/10xx/105x/peripherals.h
@@ -17,6 +17,8 @@
 #ifndef _PERIPHERALS_H_
 #define _PERIPHERALS_H_
 
+#include <board_config.h>
+
 /* Periperals configuration */
 
 /* Interrupts */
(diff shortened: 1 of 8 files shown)


hal/armv7m/imxrt/10xx/105x/peripherals.h
```diff
@@ -223,15 +225,19 @@
...
+#define FLEXSPI_COUNT 1
+#endif
 
-#define FLASH_FLEXSPI2_MOUNTED 0
+#ifndef FLEXSPI1_PORT_MASK
+#define FLEXSPI1_PORT_MASK flexspi_slBusA1
+#endif
+
+#ifndef FLEXSPI2_PORT_MASK
+#define FLEXSPI2_PORT_MASK flexspi_slBusA1
```
gerard5, line 240: On `imxrt-105x` there is only single FlexSPI instance. Add appropriate precompiler `#error` in case if someone defines more than one instance.

hal/armv7m/imxrt/117x/peripherals.h
```diff
@@ -247,20 +249,20 @@
 
 /* FLASH */
 
-#define FLASH_NO                  (FLASH_FLEXSPI1_MOUNTED + FLASH_FLEXSPI2_MOUNTED)
+#define FLASH_NO                  FLEXSPI_COUNT
```
gerard5, line 252: Previously `FLASH_FLEXSPI1_MOUNTED` and `FLASH_FLEXSPI2_MOUNTED` were used to determine which instances are used in a project. On non-standard/custom boards it is possible to have only `FLASH_FLEXSPI2_MOUNTED` instance and no `FLASH_FLEXSPI1_MOUNTED`, so if `FLEXSPI_COUNT=1` is set, you then always start with `FLEXSPI1` and not `FLEXSPI2`. Previous idea was more flexible from my point of view.

ziemleszcz: I agree that it would be nice to have this feature however I don't think it was ever there. FLASH_*_MOUNTED defines were used only to set FLASH_NO. There was never support for using FLEXSPI2 memories only.

agkaminski (approved):

PR #350, opened 2024-09-16
hal/imxrt: Fix boot from FlexSPI failing

Darchiv:
This change fixes a boot failure when plo image is larger than the hardcoded value in sflashA1Size, but smaller than m_flash size (the linker does not abort). Now (initial) flash size is tied to FlexSPI Configuration Block so that this never happens again. If plo grows too much, the linker will produce an error.

JIRA: RTOS-920

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1170-evk.

## Checklist:

## Special treatment

### DIFF
--- a/ld/armv7m7-imxrt117x.ldt
+++ b/ld/armv7m7-imxrt117x.ldt
@@ -56,8 +56,11 @@ MEMORY
 	m_dtcm    (rw)  : ORIGIN = 0x20002000 + AREA_KERNEL, LENGTH = FLEXRAM_DTCM_AREA - AREA_KERNEL
 	m_ocram1  (rwx) : ORIGIN = 0x20240000 + AREA_BOOTLOADER, LENGTH = (8 * 32k) - AREA_BOOTLOADER
 	m_ocram2  (rwx) : ORIGIN = 0x202c0000, LENGTH = 512k
-	m_flash   (rx)  : ORIGIN = 0x30000000, LENGTH = 4M
+	m_flash   (rx)  : ORIGIN = 0x30000000, LENGTH = 128k /* Not actual flash size. Initial flash size to be put into FCB block for imxrt BootROM init procedure only */
 }
+
+/* Initial flash size to be put into FCB block for imxrt BootROM init procedure only */
+_plo_flash_size = LENGTH(m_flash);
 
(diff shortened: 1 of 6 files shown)


ld/armv7m7-imxrt117x.ldt
```diff
@@ -56,9 +56,12 @@ MEMORY
 	m_dtcm    (rw)  : ORIGIN = 0x20002000 + AREA_KERNEL, LENGTH = FLEXRAM_DTCM_AREA - AREA_KERNEL
 	m_ocram1  (rwx) : ORIGIN = 0x20240000 + AREA_BOOTLOADER, LENGTH = (8 * 32k) - AREA_BOOTLOADER
 	m_ocram2  (rwx) : ORIGIN = 0x202c0000, LENGTH = 512k
-	m_flash   (rx)  : ORIGIN = 0x30000000, LENGTH = 4M
+	m_flash   (rx)  : ORIGIN = 0x30000000, LENGTH = 128k
```
nalajcie, line 59: maybe add the same comment as below also here?

ld/armv7m7-imxrt117x.ldt
```diff
@@ -56,9 +56,12 @@ MEMORY
...
 }
 
+/* initial flash size for imxrt boot procedure only */
```
nalajcie, line 62: maybe mention BootROM, eg. `initial flash size to  be put into FCB block for imxrt BootROM init procedure only`

nalajcie (dismissed):
I suggest to make the comments more descriptive, other than thad - LGTM

@ziemleszcz do you have any idea why `sflashA1Size` was previously greatly reduced in commit https://github.com/phoenix-rtos/plo/commit/44ccfa5b9ca1aa91b3513e615866a86d806a8ca6

Probably it's just for better portability - but I'm not sure.

ld/armv7m7-imxrt117x.ldt
Darchiv: Done

ld/armv7m7-imxrt117x.ldt
Darchiv: Done

PR #359, opened 2024-10-02
imxrt: fix .noxip code calling functions from .text

badochov:
JIRA: RTOS-936

## Description
imxrt-106x didn't add map entries for text, rodata, init, init_array, fini_array etc in what could have caused memory overrides if PLO was executed from RAM. (those entries got added on STM32 in https://github.com/phoenix-rtos/plo/pull/197)

imxrt used `.noxip` section for code that should be executed from XIP, however those functions called some function residing in `.text` what make `.noxip` faulty. 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/devices/flash-imx6ull/nor/nor.c
+++ b/devices/flash-imx6ull/nor/nor.c
@@ -82,7 +82,7 @@ static int nor_readID(qspi_t *qspi, u8 port, u32 *retValue, time_t timeout)
 }
 
 
-__attribute__((section(".noxip"))) int nor_readStatus(qspi_t *qspi, u8 port, u8 *statusByte, time_t timeout)
+int nor_readStatus(qspi_t *qspi, u8 port, u8 *statusByte, time_t timeout)
 {
 	struct xferOp xfer;
 
@@ -98,7 +98,7 @@ __attribute__((section(".noxip"))) int nor_readStatus(qspi_t *qspi, u8 port, u8 
 }
 
 
-__attribute__((section(".noxip"))) int nor_waitBusy(qspi_t *qspi, u8 port, time_t timeout)
+int nor_waitBusy(qspi_t *qspi, u8 port, time_t timeout)
 {
 	int res;
 	u8 status = 0;
@@ -114,7 +114,7 @@ __attribute__((section(".noxip"))) int nor_waitBusy(qspi_t *qspi, u8 port, time_
 }
 
 
-__attribute__((section(".noxip"))) int nor_writeEnable(qspi_t *qspi, u8 port, int enable, time_t timeout)
+int nor_writeEnable(qspi_t *qspi, u8 port, int enable, time_t timeout)
 {
 	struct xferOp xfer;
 	int res, seqCode;
@@ -151,7 +151,7 @@ __attribute__((section(".noxip"))) int nor_writeEnable(qspi_t *qspi, u8 port, in
 }
 
 
-__attribute__((section(".noxip"))) int nor_eraseSector(qspi_t *qspi, u8 port, addr_t addr, time_t timeout)
+int nor_eraseSector(qspi_t *qspi, u8 port, addr_t addr, time_t timeout)
 {
 	struct xferOp xfer;
 
@@ -175,7 +175,7 @@ __attribute__((section(".noxip"))) int nor_eraseSector(qspi_t *qspi, u8 port, ad
 }
 
 
-__attribute__((section(".noxip"))) int nor_eraseChip(qspi_t *qspi, u8 port, time_t timeout)
+int nor_eraseChip(qspi_t *qspi, u8 port, time_t timeout)
 {
 	struct xferOp xfer;
 
@@ -199,7 +199,7 @@ __attribute__((section(".noxip"))) int nor_eraseChip(qspi_t *qspi, u8 port, time
 }
 
 
-__attribute__((section(".noxip"))) int nor_pageProgram(qspi_t *qspi, u8 port, addr_t dstAddr, const void *src, size_t pageSz, time_t timeout)
+int nor_pageProgram(qspi_t *qspi, u8 port, addr_t dstAddr, const void *src, size_t pageSz, time_t timeout)
 {
 	struct xferOp xfer;
 
--- a/hal/armv7m/cpu.h
+++ b/hal/armv7m/cpu.h
@@ -21,28 +21,19 @@
 #include <hal/hal.h>
 
 
-static inline void hal_cpuDataMemoryBarrier(void)
-{
-	__asm__ volatile ("dmb");
-}
+/* NOTE: defined as macros to make sure there are no calls to code in flash in .noxip sections. */
+/* clang-format off */
+#define hal_cpuDataMemoryBarrier() do { __asm__ volatile ("dmb"); } while(0)
 
 
-static inline void hal_cpuDataSyncBarrier(void)
-{
-	__asm__ volatile ("dsb");
-}
+#define hal_cpuDataSyncBarrier() do { __asm__ volatile ("dsb"); } while(0)
 
 
-static inline void hal_cpuInstrBarrier(void)
-{
-	__asm__ volatile ("isb");
-}
+#define hal_cpuInstrBarrier() do { __asm__ volatile ("isb"); } while(0)
 
 
-static inline void hal_cpuHalt(void)
-{
-	__asm__ volatile ("wfi");
-}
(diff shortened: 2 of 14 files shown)


badochov:
I'll add JIRA issue and split this into two PRs one for noxip one for map once I get some feedback that those changes are sane

badochov:
Demo with -fno-strict-aliasing, that fails without those changes
https://github.com/phoenix-rtos/phoenix-rtos-project/actions/runs/11147095089/job/30980943887?pr=1190

damianloew:
For now tested ~30 times and the mentioned issue hasn't been caught when before it was seen ~4/20 flash attempts.

jmaksymowicz:
Looks okay, from my tests I haven't noticed any crashes

lib/cbuffer.h
```diff
@@ -40,6 +40,12 @@ extern size_t lib_cbufRead(cbuffer_t *buf, void *data, size_t sz);
 extern size_t lib_cbufWrite(cbuffer_t *buf, const void *data, size_t sz);
 
 
+extern int lib_cbufReadByte(cbuffer_t *buf, char *data) __attribute__((section(".noxip")));
```
agkaminski, line 43: Other function prototypes does not have the attribute - is it needed or not (I think not)?

hal/armv7m/imxrt/10xx/console.c
```diff
@@ -76,7 +76,7 @@ void hal_consoleSetHooks(ssize_t (*writeHook)(int, const void *, size_t))
 }
 
 
-void hal_consolePrint(const char *s)
+__attribute__((section(".noxip"))) void hal_consolePrint(const char *s)
```
agkaminski, line 79: Duplicate the change in `117x` too

hal/armv7m/cpu.h
```diff
@@ -21,28 +21,19 @@
 #include <hal/hal.h>
 
 
-static inline void hal_cpuDataMemoryBarrier(void)
-{
-	__asm__ volatile ("dmb");
-}
+/* NOTE: defined as macros to make sure there are no calls to code in flash in .noxip sections. */
+/* clang-format off */
+#define hal_cpuDataMemoryBarrier() do { __asm__ volatile ("dmb"); } while(0)
```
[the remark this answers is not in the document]
badochov: `inline.c`
```c
#include "inline.h"

int main(void) {
    return test();
}
```

`inline.h`
```c
static inline int test(void) {
    for (int i =0 ;i < 4096;i++);
    return 0;
}
```

`gcc -O0 inline.c`


`readelf -s`
```
    12: 0000000000001129    33 FUNC    LOCAL  DEFAULT   14 test
```

`objdump -d`

```
0000000000001129 <test>:
    1129:	55                   	push   %rbp
    112a:	48 89 e5             	mov    %rsp,%rbp
    112d:	c7 45 fc 00 00 00 00 	movl   $0x0,-0x4(%rbp)
    1134:	eb 04                	jmp    113a <test+0x11>
    1136:	83 45 fc 01          	addl   $0x1,-0x4(%rbp)
    113a:	81 7d fc ff 0f 00 00 	cmpl   $0xfff,-0x4(%rbp)
    1141:	7e f3                	jle    1136 <test+0xd>
    1143:	b8 00 00 00 00       	mov    $0x0,%eax
    1148:	5d                   	pop    %rbp
    1149:	c3                   	ret

000000000000114a <main>:
    114a:	f3 0f 1e fa          	endbr64
    114e:	55                   	push   %rbp
    114f:	48 89 e5             	mov    %rsp,%rbp
    1152:	e8 d2 ff ff ff       	call   1129 <test>
    1157:	5d                   	pop    %rbp
    1158:	c3                   	ret
```

----------------------------------------------------

In this case symbol is created, and I see no reason why each compilation unit can't have `static inline` function defined under some alias if the function is big enough.
In practice though GCC on -O2 should always inline a one liner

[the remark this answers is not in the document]
agkaminski: Ok, you convinced me ;)

lib/cbuffer.h
agkaminski: Ack

agkaminski:
LGTM

hal/armv7m/cpu.h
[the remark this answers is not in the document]
nalajcie: AFAIK `-O0` means `-fno-inline` so we explicitly disable inlining for functions defined as `inline` :)
image

Inlining happens with `-O`, `-O1`, `-Os` and `-Og`.

IMHO for single-instruction intrinsics it's ok to define them as macros, for more complex implementations (eg. `clz` on some archs) I would go with function marked as `always_inline`

hal/armv7m/string.c
```diff
@@ -240,7 +240,7 @@ char *hal_strchr(const char *s, int c)
 }
 
 
-int hal_i2s(char *prefix, char *s, unsigned int i, unsigned char b, char zero)
+__attribute__((section(".noxip"))) int hal_i2s(char *prefix, char *s, unsigned int i, unsigned char b, char zero)
```
nalajcie, line 243: side discussion: maybe we should switch to 
```c
__attribute__((section(".noxip"))) 
int hal_i2s(char *prefix, char *s, unsigned int i, unsigned char b, char zero)
{
    // ...
}
```
style?

devices/flash-imx6ull/nor/nor.c
```diff
@@ -82,7 +82,7 @@ static int nor_readID(qspi_t *qspi, u8 port, u32 *retValue, time_t timeout)
 }
 
 
-__attribute__((section(".noxip"))) int nor_readStatus(qspi_t *qspi, u8 port, u8 *statusByte, time_t timeout)
+int nor_readStatus(qspi_t *qspi, u8 port, u8 *statusByte, time_t timeout)
```
nalajcie, line 85: not sure: should developer know we're not using XIP on 6ull or should we always mark all functions used in flash access as `noxip`? (eg. XIP is possible on `zynq7000`, currently we're not supporting/using it)

badochov: In my opinion we either should provide XIP support fully or not at all, being in the middle causes biggest confusion. If `.noxip` was to be used on `imx6ull` then we should also mark all interrupt handlers as `.noxip` and all function called by those function.

hal/armv7m/string.c
badochov: The approach you proposed is more readable while browsing a single file, but when searching for definitions and checking if they are `.noxip` it introduces much more friction. However, I don't think that after this PR looking for `.noxip` usages is going to be a common problem.

badochov:
Map entries PR: https://github.com/phoenix-rtos/plo/pull/360

badochov:
imxrt106x - flash failed after removing map entires change from this PR, thus those were the root cause of this problem.

`.noxip` made code very finicky to size and alignment changes
https://github.com/phoenix-rtos/plo/actions/runs/11159904407/job/31019379132?pr=359

devices/flash-imx6ull/nor/nor.c
nalajcie: Yeah, for me the easiest for developer would be to have a single rule eg. "mark all code used in interruptions and NVM storage as noxip" (without the necessity to know whether we can/currently use XIP on that platform or some more confusing rules (eg. common code for arm7a needs to be noxip because of zynq, but imx6ull-specific code can be all xip (as XIP is not supported by this CPU))).

The decision to run from flash or  from RAM is then done only in linker script (right now we do it as `plo` and `plo-ram`, but might make per-project ld scripts in the future).

This is obviously a separate discussion from this PR, just voicing my thoughts.

hal/armv7m/string.c
nalajcie: I don't think it's harder (`rg -A1 noxip`), but it depends on the tool i suppose ;)

nalajcie: ... buuut the `clang-format` cannot be currently configured to put a newline after the attribute - I see that there is a stale PR open which adds the functionality (https://github.com/llvm/llvm-project/pull/78102), so it will take some time to be backed up into stable release of clang-format :)

devices/flash-imx6ull/nor/nor.c
badochov: Also it would be great to have some static analysis tools that would catch `.noxip` code calling `.text`

agkaminski (approved):

jmaksymowicz (approved):

PR #365, opened 2024-10-22
hal/armv7m: allow for more granular memory maps

jmaksymowicz:
Use MPU region overlap feature to allow for more granularity when translating memory maps to MPU regions.

## Description
On armv7m, the MPU only supports regions if their size is a power of 2 and they are aligned to their size. Additionally, each region is divided into 8 equally sized subregions which can be enabled individually. Memory maps passed using `map` command must be translated into one or more MPU regions so that memory protection can be enforced.

In practice this means that the start and end address representable with one region can have up to 3 "significant bits" - for example, an address range `0x20000 - 0x20500` can be represented with a single region, but range `0x20000 - 0x20900` requires 2 regions. Currently our code will limit assigning regions to 2 per map, which means that attempting to create a more granular memory map will fail.

To work around this without creating excessively many regions, the new code uses the fact that armv7m MPU supports overlapping regions. It detects how many regions would be necessary, if more than 2 the code will attempt to allocate a region larger than the memory map (rounded up to smallest representable region) then block off the excess using another memory region. This "blocking" region is set to not allow any access to the "blocked" memory. Due to the way priorities work, any subsequent mapping that overlaps this "blocking" region will take precedence.

However, if a "blocking" region were to overlap an existing region, it would take precedence and MPU will not allow access even if a program is allowed access to both memory maps. To prevent this unintuitive behavior, the code checks for this situation and fails the `map` command. This has a side effect that now the order in which `map` commands are issued may affect their success or failure. This isn't optimal, but thanks to failing early these situations would be easy to detect and fix.

For example: based on the following memory maps
```
map dtcm 0x20000000 0x2003f000 rwcb
map rtt 0x2003f000 0x20040000 rw
```
the new code will create MPU regions
```
MAP NAME  REGION  SUB  START       END         EN  XN  PERM P/U  TEX  S  C  B
dtcm      1       all  0x20000000  0x2003ffff  y   y   rw/rw     000  n  y  y
dtcm      2       all  0x2003f000  0x2003ffff  y   n   na/na     000  n  n  n
rtt       3       all  0x2003f000  0x2003ffff  y   y   rw/rw     000  n  n  n
```

## Motivation and Context
For using RTT a memory map needs to be created, which most likely means cutting out a part of an existing memory map. This PR allows the "cutout" to be as small as possible.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt117x

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/mpu.c
+++ b/hal/armv7m/mpu.c
@@ -20,99 +20,190 @@
 static mpu_common_t mpu_common;
 
 
+/* clang-format off */
 enum { mpu_type, mpu_ctrl, mpu_rnr, mpu_rbar, mpu_rasr, mpu_rbar_a1, mpu_rasr_a1, mpu_rbar_a2, mpu_rasr_a2,
 	   mpu_rbar_a3, mpu_rasr_a3 };
+/* clang-format on */
 
 
-/* Binary count trailing zeros */
-__attribute__((always_inline)) static inline u8 _ctz(u32 val)
-{
-	asm volatile(" \
-		rbit %0, %0; \
-		clz %0, %0"
-		: "+l"(val));
-	return val;
-}
+/* Removes all RASR attribute bits except ENABLE */
+#define HOLE_ATTR(rasrAttr) (0 | ((rasrAttr) & 0x1))
 
 
 /* Setup single MPU region entry in a local MPU context */
-static int mpu_regionSet(u8 idx, addr_t addr, u8 sizeBit, u32 attr, unsigned int enable, u8 srdMask)
+static int mpu_regionSet(unsigned int *idx, u32 baseAddr, u8 srdMask, u8 sizeBit, u32 rasrAttr)
 {
-	u8 tex = 0;
-	u8 ap = 1; /* privileged read-write access, unprivileged no access */
-
-	if (sizeBit < 5 || idx >= mpu_common.regMax)
+	if ((sizeBit < 5) || (*idx >= mpu_common.regMax)) {
 		return -EPERM;
+	}
 
-	if (attr & mAttrRead)
-		ap = 2; /* privileged read-write access, unprivileged read only access */
+	mpu_common.region[*idx].rbar =
+			baseAddr |
+			(1u << 4) | /* mark region as valid */
+			(*idx & 0xfu);
 
-	if (attr & mAttrWrite)
-		ap = 3; /* privileged read-write access, unprivileged read and write access */
+	mpu_common.region[*idx].rasr =
+			rasrAttr |
+			(((u32)srdMask) << 8) |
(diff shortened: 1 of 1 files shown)


hal/armv7m/mpu.c
```diff
@@ -20,99 +20,181 @@
...
 
 
-/* Find best alignment for map and setup a region */
-static int mpu_regionBestFit(u8 idx, addr_t addr, size_t size, u32 attr, unsigned int enable, size_t *allocSize)
+static int mpu_checkOverlap(unsigned int idx, u32 start, u32 end)
 {
-	size_t srSize;
-	addr_t srBase, alignMask;
+	end -= 1;
+	for (int i = 0; i < idx; i++) {
```
agkaminski, line 82: Not allowed in kernel and plo, please move variables declarations to the beginning of the function

hal/armv7m/mpu.c
```diff
@@ -20,99 +20,181 @@
...
+	/* Allow end == 0, this means end of address range */
+	if ((end != 0) && (end <= start)) {
+		return -EINVAL;
+	}
 
-	if (alignBits < 5) {
-		*allocSize = 1u << alignBits;
+	/* Check if size is power of 2 and start is aligned - necessary for handling
+	   small regions (below 256 bytes) */
+	const u32 size = (end - start) & 0xfffffffful;
```
agkaminski, line 129:
```suggestion
	const size_t size = (end - start) & 0xfffffffful;
```

jmaksymowicz: Done

hal/armv7m/mpu.c
```diff
@@ -20,99 +20,181 @@
...
 
-/* Find best alignment for map and setup a region */
-static int mpu_regionBestFit(u8 idx, addr_t addr, size_t size, u32 attr, unsigned int enable, size_t *allocSize)
+static int mpu_checkOverlap(unsigned int idx, u32 start, u32 end)
 {
-	size_t srSize;
-	addr_t srBase, alignMask;
+	end -= 1;
+	for (int i = 0; i < idx; i++) {
+		if (((mpu_common.region[i].rbar & 0x10ul) == 0) || ((mpu_common.region[i].rasr & 0x1ul) == 0)) {
```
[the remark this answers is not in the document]
jmaksymowicz: Done

hal/armv7m/mpu.c
jmaksymowicz: OK, moved `const` declarations to the start as well, but I'm not sure if it was necessary.

agkaminski (approved):

Darchiv (approved):

PR #367, opened 2024-11-05
hal/imxrt117x: fix IOpad

julianuziemblo:
## Description
- properly set ranges of registers configured differently
- add a way of setting ODE field in non-LPSR/SNVS GPIO registers
- set proper ranges for setting ODE field in LPSR, SNVS, and remaining GPIO registers

## Motivation and Context
related: https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/618

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `armv7m7-imxrt1170-evk`.

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/imxrt/117x/imxrt.c
+++ b/hal/armv7m/imxrt/117x/imxrt.c
@@ -158,8 +158,7 @@ __attribute__((section(".noxip"))) int _imxrt_setIOpad(int pad, char sre, char d
 		return -1;
 	}
 
-	if (((pad >= pctl_pad_gpio_emc_b1_00) && (pad <= pctl_pad_gpio_emc_b2_20)) ||
-			((pad >= pctl_pad_gpio_sd_b1_00) && (pad <= pctl_pad_gpio_disp_b2_15))) {
+	if ((pad <= pctl_pad_gpio_emc_b2_20) || ((pad >= pctl_pad_gpio_sd_b1_00) && (pad <= pctl_pad_gpio_disp_b1_11))) {
 		/* Fields have slightly diffrent meaning... */
 		if (pue == 0) {
 			pull = 3;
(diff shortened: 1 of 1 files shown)


hal/armv7m/imxrt/117x/imxrt.c
```diff
@@ -178,11 +178,15 @@ __attribute__((section(".noxip"))) int _imxrt_setIOpad(int pad, char sre, char d
...
+		}
+		else if ((pad >= pctl_pad_wakeup) && (pad <= pctl_pad_gpio_snvs_09)) {
 			t &= ~(1 << 6);
 			t |= !!ode << 6;
 		}
-		else {
+		else if ((pad >= pctl_pad_gpio_lpsr_00) && (pad <= pctl_pad_gpio_lpsr_15)) {
 			t &= ~(1 << 5);
 			t |= !!ode << 5;
 		}
```
agkaminski, line 192: `else` error perhaps?

agkaminski: Range check is performed in `_imxrt_IOpadGetReg`, perhaps it's ok to replace last `else if` with `else`? Or does some range of pads does not allow `ode` setting?

julianuziemblo: `pctl_pad_test_mode`, `pctl_pad_por_b` and `pctl_pad_onoff` pads' ODE field is read-only

julianuziemblo: I think that error on else after those checks would be a good idea

agkaminski: Ok. Sadly, we still need an empty `else` because of MISRA

agkaminski: > I think that error on else after those checks would be a good idea

Perhaps I don't understand, but if we can't set `ode` it's ok to just skip it, if we error on `else`, then we will always error on pads that do not have `ode` setting

julianuziemblo: okay, that's valid, I forgot about that. I will try to rewrite it somehow

agkaminski (approved):

PR #419, opened 2026-03-06
Introduce Kernel Partitioning

etiaro:
## Description

Add loader-side support for kernel partitioning.

Key changes:
- Introduce `part` command for creating partitions with configurable memory maps (allocable and accessible) and flags (e.g. IPC permission)
- Introduce `sched` command for scheduler temporal windows configuration
- Rework `mpu` command to show per-app MPU configuration
- Move MPU configuration originally somewhat dynamicly created inside kernel to plo-created `syspage_part_t`
- Add optional partition name argument to `app` command
backward-compatible default partition is created by each app without partition specified.

## Motivation and Context

This is the companion change to the kernel partitioning support. Currently system configuration mechanisms are based on plo-scripts, so partitioning configuration follows that convetion.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/cmds/mpu.c
+++ b/cmds/mpu.c
@@ -82,54 +82,58 @@ static void mpu_regionPrint(const char *name, u32 rbar, u32 rasr)
 
 static void cmd_mpuInfo(void)
 {
-	lib_printf("prints the use of MPU regions, usage: mpu [all]");
+	lib_printf("prints the use of MPU regions, usage: mpu [prog name]");
 }
 
 
 static int cmd_mpu(int argc, char *argv[])
 {
 	const char *name;
-	unsigned int i, regCnt;
-	const mpu_region_t *region;
-	const mpu_common_t *const mpu_common = mpu_getCommon();
+	unsigned int i;
+	const syspage_prog_t *prog = syspage_progsGet();
+	const syspage_prog_t *firstProg = prog;
+	const unsigned int regMax = mpu_getMaxRegionsCount();
 
-	if (argc == 1) {
-		regCnt = mpu_common->regCnt;
+	if (prog == NULL) {
+		log_error("\nNo programs in syspage!");
+		return CMD_EXIT_FAILURE;
 	}
-	else if (argc == 2) {
-		if (hal_strcmp(argv[1], "all") != 0) {
-			log_error("\n%s: Wrong arguments", argv[0]);
-			return CMD_EXIT_FAILURE;
-		}
 
-		regCnt = mpu_common->regMax;
-	}
-	else {
+	if (argc > 2) {
 		log_error("\n%s: Wrong argument count", argv[0]);
 		return CMD_EXIT_FAILURE;
 	}
 
-	if (mpu_common->regMax != sizeof(((hal_syspage_t *)0)->mpu.table) / sizeof(((hal_syspage_t *)0)->mpu.table[0])) {
-		log_error("\n%s: MPU hal is not initialized or unsupported type was detected", argv[0]);
-		return CMD_EXIT_FAILURE;
-	}
-
-	lib_printf(CONSOLE_BOLD "\n%-9s %-7s %-4s %-11s %-11s %-3s %-3s %-9s %-4s %-2s %-2s %-2s\n" CONSOLE_NORMAL,
-		"MAP NAME", "REGION", "SUB", "START", "END", "EN", "XN", "PERM P/U", "TEX", "S", "C", "B");
-
-	for (i = 0; i < regCnt; i++) {
-		region = &mpu_common->region[i];
-		name = syspage_mapName(mpu_common->mapId[i]);
-
-		if (name == NULL) {
-			name = "<none>";
+	do {
+		name = (prog->argv[0] == 'X') ? prog->argv + 1 : prog->argv;
+		if ((argc == 2) && (hal_strcmp(name, argv[1]) != 0)) {
+			prog = prog->next;
+			continue;
+		}
+		lib_printf("\n%-16s 0x%08x%4s 0x%08x", name, prog->start, "", prog->end);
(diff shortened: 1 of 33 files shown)


cmds/part.c
```diff
@@ -0,0 +1,192 @@
...
+{
+	int res, i;
+
+	*schedIds = 0;
+
+	for (i = 0; i < nb; ++i) {
+		res = lib_strtoul(schedStr, &schedStr, 10);
+
+
+		if (res >= syspage_schedulerWindowCount()) {
```
adamgreloch:
```suggestion
		res = lib_strtoul(schedStr, &schedStr, 10);
		if (res >= syspage_schedulerWindowCount()) {
```

cmds/mpu.c
```diff
@@ -82,54 +82,58 @@ static void mpu_regionPrint(const char *name, u32 rbar, u32 rasr)
...
-		log_error("\n%s: Wrong argument count", argv[0]);
+	if (part == NULL) {
+		log_error("\nNo partitions in syspage!");
 		return CMD_EXIT_FAILURE;
 	}
 
-	if (mpu_common->regMax != sizeof(((hal_syspage_t *)0)->mpu.table) / sizeof(((hal_syspage_t *)0)->mpu.table[0])) {
-		log_error("\n%s: MPU hal is not initialized or unsupported type was detected", argv[0]);
+	if (argc > 2) {
+		log_error("\n%s: Wrong argument count", argv[0]);
```
adamgreloch: `Too many arguments`

ziemleszcz:
Please rebase this PR.

etiaro:
Rebasing with feature/ branch (and all other breaking submodules)

etiaro:
Changed base branch to avoid problems with branch protection while rebasing and CI errors spam.

At the same time, changed the implementation from by default creating partition-per-app to single default partition, as it seems like a cleaner for the future features.

Introduced Per-CPU scheduler configuration for improved elasticity

Removed NOMMU multimap allocation, as it was only partially implemented and broke with the new conception of default partition. Will be introduced in the future together with MMU NUMA support.

cmds/part.c
```diff
@@ -0,0 +1,128 @@
...
+		log_error("\nCannot allocate memory for %s", name);
+		return -ENOMEM;
+	}
+
+	part->schedWindow = lib_strtoul(schedWindow, &schedWindow, 10);
+	if (*schedWindow != '\0') {
+		log_error("\n%s: Invalid arguments", argv[0]);
+		return -EINVAL;
+	}
+	if (part->schedWindow > syspage_schedulerConfigGet()->windowCnt) {
```
ziemleszcz: Potential NULL dereference on `syspage_schedulerConfigGet()`?

cmds/part.c
```diff
@@ -0,0 +1,128 @@
...
+
+	/* ARG_3: scheduler windows */
+	schedWindow = argv[argvID++];
+
+	/* ARG_4: memory allocation limit */
+	availableMem = argv[argvID];
+
+	accessSz = cmd_listParse(accessMaps, ';');
+
+	part = syspage_partAdd();
```
ziemleszcz: Consider doing this as the final step after all allocations and parsing succeed. Use local variables for intermediate results and assign the fields only after everything succeeds.

cmds/sched.c
```diff
@@ -0,0 +1,147 @@
...
+		if (cycle->windows[i].id > windowCnt) {
+			log_error("\nInvalid scheduler cycle description: window ID exceeds window count");
+			return NULL;
+		}
+		if (*cur != ':') {
+			log_error("\nInvalid scheduler cycle description");
+			return NULL;
+		}
+		duration += lib_strtoul(cur + 1, &cur, 0);
+		cycle->windows[i].stop = duration;
```
ziemleszcz, line 92: Is duration=0 valid here?

etiaro: No, its not. I'll introduce a check (although small values like 1 make no sense either).

cmds/part.c
etiaro: I need to rethink the default scheduler config that is currently created inside first `app` call. Thanks.

etiaro:
For some reason PLO grows >10kB on sparc target. Seems suspicious, as e.g. 7a9dc97 only introduces new functions while reducing this binary size back to reasonable values.
Needs further investigation. 

EDIT: for now just increased kernel offset in sparc config

syspage.c
```diff
@@ -482,6 +521,99 @@ const char *syspage_mapName(u8 id)
...
+
+	log_info("\nUsing default scheduler configuration");
+	return config;
+}
+
+
+syspage_sched_t *syspage_schedulerConfigGet(void)
+{
+	if (syspage_common.syspage->sched == NULL) {
+		syspage_common.syspage->sched = syspage_schedulerCreateDefault();
```
ziemleszcz: I don't like this approach. Get function should not set. Maybe better approach is to start with the default scheduler and allow set to overwrite it (but only the default one).

etiaro: That was my first idea, but it will lead to wasted syspage space for the default configuration when another one is installed

ziemleszcz: Can't we reuse default scheduler in sched (get it, modify and set)?

etiaro: Sched config relies on Flexible array members to reduce dereference count, increase cache locality and therefore scheduler overhead. I don't have the actual numbers to tell if the difference is meaningful at all.

Even with regular pointer it can't be fully reused as lengths of arrays depend on particular config.
For now I see three options:
- waste some syspage memory
- set some fixed MAX number of windows in a cycle and always allocate full array
- leave it as it is now

I'll think of some other ways, for now 1st one seems the most reasonable as it don't seem to be a large quantity anyway (~20-30 b), and we waste some memory on alignments anyway

